PR #1, opened 2017-05-19
Move code for deploying OpenWhisk on Kube from apache/incubator-openwhisk-devtools

DanLavine:
* Update any file path references for openwhisk-devtools to incubator-openwhisk-deploy-kube


mrutkows:
Thanks for moving here from https://github.com/apache/incubator-openwhisk-devtools

PR #2, opened 2017-05-25
Build kube 1.5.6 and 1.6.2 images and update ci

DanLavine:
* Create a kube 1.6.2 deployment image.
* Update CI to test both kube versions
* Update Docker scripts to build all images

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,12 @@ sudo: required
 
 env:
   matrix:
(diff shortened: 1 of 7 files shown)


mrutkows:
It appears that from the changes that primarily "etcd" version and kube version are linked/co-dependent.  Also, altering the travis config to use env. vars for versions of both seems to be more flexible (and proper).

PR #4, opened 2017-06-01
Allow more time for Kafka to start

tobias:
Locally with a fresh environment (with no docker images downloaded) the
default loop count isn't enough to ensure that the kafka pod is fully
started. I have signed the Apache CLA as "Tobias Oliver Crawley".

Issue #9, opened 2017-06-06
CI occasionally flakes out


DanLavine:
Every once and a while CI will fail with:

```
error: error validating "configure/openwhisk_kube_namespace.yml": error validating data: open /home/travis/.kube/schema/v1.5.6/api/v1/schema.json: permission denied; if you choose to ignore these errors, turn validation off with --validate=false
```

For whatever reason we cannot open the file `/home/travis/.kube/schema/v1.5.6/api/v1/schema.json` because we do not have permissions. Initially this file is created via a `sudo` command when setting up Kubernetes, but I thought that I had put in a fix for this [here](https://github.com/apache/incubator-openwhisk-deploy-kube/blob/master/.travis/setup.sh#L56-L60)

DanLavine:
Just had a probable PR fix merged. Will monitor it to make sure that it stops going red for file permissions.

DanLavine:
Think I am going to close this issue. Does not appear to be happening anymore.

PR #13, opened 2017-06-12
Add script and instructions for using Minikube

tobias:
This adds adds experimental support for supporting Minikube. It includes a script that configures Minikube's `docker0` interface, starts the configure job, and uses kube's patch functionality to modify the environment for the invoker StatefulSet to downgrade the Docker client's API version to work with the older Docker daemon in Minikube.

This also includes a pair of scripts in `ci/` (modeled after the existing scripts in `/.travis`) that can be used as the basis of a CI setup if CI is run on a platform that allows Minikube to create VMs. 

Without #4, the `deploy_minikube.sh` script fails on the first run on a new Minikube install due to a timeout waiting for Kafka to start, but succeeds on a second run (after calling `/configure/cleanup.sh`). Once #4 is merged, this should work more reliably on a fresh install.

Let me know if you need anything else here or have any questions.

### DIFF
--- a/minikube/README.md
+++ b/minikube/README.md
@@ -1,0 +1,41 @@
+# Deploying OpenWhisk to Minikube
+
+**Note:** this is currently experimental, and the integration tests
+don't yet run against Minikube on CI. Minikube currently uses an old
+version of Docker (v1.11.x), so may not be fully compatible with
+OpenWhisk (see [Requirements](../README.md#requirements) for more
+info). These steps have only been manually tested with Minikube
+v0.19.1.
+
+
+# Installation
+
+1. Download and set up [Minikube](https://github.com/kubernetes/minikube)
+2. Start minikube with `minikube start`
+3. run `./deploy_minikube.sh`. This will configure Minikube for
+   OpenWhisk, run the standard Kubernetes OpenWhisk deployment, then
+   reconfigure the Invoker so it can properly talk to the Docker
+   daemon used by Minikube
+
+Once you've started the deploy script, you can follow along on the
+progress of the configuration task as noted in the
+[top-level README](../README.md#configure-openwhisk). 
+
+
+# The details
+
+There are currently two changes required in order for OpenWhisk to
+work on Minikube, both of which are performed by the deploy script:
+
+1. The `docker0` interface inside Minikube needs to be put into
+   promiscuous mode to allow Kafka to talk to ZooKeeper
+2. The Docker client used by the Invoker is newer than the Docker
+   daemon provided by Minikube, so we have to set an environment
+   variable (`DOCKER_API_VERSION`) in the environment for the
+   Invoker's container (via the StatefulSet template) to force it to
+   use an older version of the API protocol (1.23). We then need to
+   delete the existing Invoker pod that was started by the initial
+   deploy to force a new one to be created with the new environment.
+
+
+
--- a/minikube/ci/setup.sh
+++ b/minikube/ci/setup.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+# Installs and starts minikube
+
+set -x
+
+# download kubectl
+wget https://storage.googleapis.com/kubernetes-release/release/v1.5.4/bin/linux/amd64/kubectl
+chmod +x kubectl
+sudo mv kubectl /usr/local/bin/kubectl
+
+# download minikube
+wget https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
+chmod +x minikube-linux-amd64
+sudo mv minikube-linux-amd64 /usr/local/bin/minikube
+minikube version
+
+# assumes virtualbox, override with  --vm-driver virtualbox|kvm|xhyve|vmwarefusion
+minikube start
+
+# Wait until minikube is up and running
+TIMEOUT=0
+TIMEOUT_COUNT=40
+until $(minikube status &> /dev/null) || [ $TIMEOUT -eq $TIMEOUT_COUNT ]; do
+  echo "Minikube is not up yet"
+  let TIMEOUT=TIMEOUT+1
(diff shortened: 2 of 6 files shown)


minikube/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+
+# Installation
+
+1. Download set up [Minikube](https://github.com/kubernetes/minikube)
+2. Start minikube with `minikube start`
+3. run `./deploy_minikube.sh`. This will configure Minikube for
+   OpenWhisk, run the standard Kubernetes OpenWhisk deployment, then
+   reconfigure the Invoker so it can properly talk to the Docker
+   daemon used by Minikybe
```
mlangbehn: This line should be:

```markdown
  daemon used by Minikube
```

minikube/deploy_minikube.sh
```diff
@@ -0,0 +1,60 @@
...
+  fi
+  printf "."
+  let TIMEOUT=TIMEOUT+1
+  sleep 30
+done
+echo
+
+if [ "$PASSED" = false ]; then
+  echo "The job to configure OpenWhisk did not finish successfully"
+  echo "Execute `kubectl -n openwhisk logs $CONFIGURE_POD` to see the output from the configure job"
```
mlangbehn: Escape the backticks so `kubectl` doesn't actually run here:

```shell
  echo "Execute \`kubectl -n openwhisk logs $CONFIGURE_POD\` to see the output from the configure job"
```

tobias:
@mlangbehn: hanks for catching those issues - I just pushed a new commit with those items fixed.

minikube/ci/setup.sh
```diff
@@ -0,0 +1,38 @@
...
+chmod +x kubectl
+sudo mv kubectl /usr/local/bin/kubectl
+
+# download minikube
+wget https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
+chmod +x minikube-linux-amd64
+sudo mv minikube-linux-amd64 /usr/local/bin/minikube
+minikube version
+
+# assumes virtualbox, override with  --vm-driver virtualbox|xhyve|vmwarefusion
```
mlangbehn: We should include KVM in this list:

```shell
# assumes virtualbox, override with  --vm-driver virtualbox|kvm|xhyve|vmwarefusion
```

minikube/deploy_minikube.sh
```diff
@@ -0,0 +1,60 @@
...
+
+echo "Deleting existing invoker pod so it will get recreated with new env"
+kubectl -n openwhisk delete po/invoker-0
+
+CONFIGURE_POD=$(kubectl get pods --all-namespaces -o wide | grep configure | awk '{print $2}')
+
+printf "Waiting for the configure job to complete"
+PASSED=false
+TIMEOUT=0
+until $PASSED || [ $TIMEOUT -eq 25 ]; do
```
mlangbehn: Should we use `TIMEOUT_COUNT` here for uniformity?

tobias: Good call, updated.

minikube/ci/setup.sh
tobias: Added.

tobias:
That TODO is originally from `.travis/builds.sh`, since `ci/build.sh` is a modified copy of that file, so I left it in place. I think it would be worthwhile to modify the existing travis scripts with conditional logic for testing on minikube once that is an option (or extracting the commonality out to other scripts if an additional CI platform is adopted to support minikube testing).

minikube/deploy_minikube.sh
```diff
@@ -0,0 +1,61 @@
...
+
+echo "Deleting existing invoker pod so it will get recreated with new env"
+kubectl -n openwhisk delete po/invoker-0
+
+CONFIGURE_POD=$(kubectl get pods --all-namespaces -o wide | grep configure | awk '{print $2}')
+
+printf "Waiting for the configure job to complete"
+PASSED=false
+TIMEOUT=0
+TIMEOUT_COUNT=25
```
mlangbehn: On my test VM (Ubuntu 16.04, 16 GB RAM, 8x 2.6 GHz CPU), this timeout wasn't long enough. Switching `TIMEOUT_COUNT` to 500 was also not long enough. Running `kubectl -n openwhisk logs [configure-job]` shows an ansible failure on either consul or couchdb (play and task appears to be inconsistent).

@tobias, are you seeing any similar issues?

tobias: @mlangbehn No, I'm not seeing any issues with consul or couchdb. The only issue I have is the kafka timeout that should go away once #4 makes it into the next build of the configure docker image. What VirtualBox driver are you using? And can you share the log from the configure job?

mlangbehn: @tobias: If you aren't seeing any issues in your testing, then it's probably an issue with my setup. I am running minikube with KVM, so I will debug on my side.

tobias:
@mrutkows The Travis failures don't appear to be related to these changes, it looks like they are the same issue as #9. Travis currently isn't verifying the Minikube support at all, since we can't run Minikube on Travis. I have noted that the minikube is experimental anywhere it is mentioned in the READMEs, is the wording there sufficient? I also note in `minikube/README.md` that it isn't currently verified by CI.

DanLavine:
Yeah. Thats what these ci failures look like. Moving forward, I think that there is room to experiment in how OpenWhisk is deployed to Kubernetes. With the intent to get away from the required configuration image and have everything built into the docker images as much as possible i think there is more we can do with CI. For example, one CI would be to put in the invoker properties to downgrade the docker API version and make sure that everything works.

However that is work for the future we need to keep on top of. Currently this should be fine.

mrutkows:
Merging minikube work as it does impact other deployments and is separated.

PR #43, opened 2017-07-25
WIP: configure/deploy via helm charts

berndtj:
This commit is an example of using charts to deploy each
openwhisk component via charts instead of managing with ansible.

Taken a little bit further the entire application could be
deployable using a single chart (which references other charts).
All of the logic in the ansible scripts can be constructed with
kubernetes and helm functionality.

For instance deploying nginx is now:

	helm install --name demo --set nginx.domainName=localhost charts/nginx

The certs are generated by an init container automatically.

This is a WIP.  I wanted to open it up for review/comments.

### DIFF
--- a/charts/controller/.helmignore
+++ b/charts/controller/.helmignore
@@ -1,0 +1,21 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
--- a/charts/controller/Chart.yaml
+++ b/charts/controller/Chart.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+description: A Helm chart for Kubernetes
+name: controller
+version: 0.1.0
--- a/charts/controller/templates/NOTES.txt
+++ b/charts/controller/templates/NOTES.txt
@@ -1,0 +1,4 @@
+1. Get the application URL by running these commands:
+  export POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app={{ template "name" . }},release={{ .Release.Name }}" -o jsonpath="{.items[0].metadata.name}")
+  echo "Visit http://127.0.0.1:8080 to use your application"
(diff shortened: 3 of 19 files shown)


ddragosd:
@berndtj IMO this looks very interesting. I think it makes sense to use a tool more _native_ to Kube to deploy OpenWhisk. 
For Mesos we're following a similar approach, reusing [DC/OS Universe](https://github.com/mesosphere/universe).

I'm not familiar with `helm` but I'd love to see it in action with OpenWhisk once this PR is complete.

DanLavine:
There was some communication in the OpenWhisk Kube channel and so I;m going to capture the jist of that discussion for reference:

This PR seems like a great idea that I think everyone would benefit from. It solves the issues where a lot of properties/environment variables need to be defined correctly all over the place. 

Sadly I think that there is a bigger issue before this can be full and easily implemented. 
What I think we first need is one yaml file that defines all the values users need to define/have defaults if users do not define those values. Then we need a consistent way to populate that yaml file with environment deployment properties. This way that one yaml file can feed into deployment option like: Kube, Ansible, Docker-Compose, mesos, etc.

xingzhou:
just created another ticket in k8s official charts repo:
https://github.com/kubernetes/charts/issues/1648

Once this is stable, can we also submit a PR in kubernetes charts repo?

xingzhou:
@berndtj, are you still working on this? as we are also trying to write a chart to deploy open whisk, not sure if we can cooperate together?

berndtj:
I/we are working on it.  I was on PTO for the past couple weeks, and I delegated to a colleague.  That said, I'm sure he would like some help with it.  I'm trying to get his github ID so that you can connect.

I also chatted with @DanLavine on slack yesterday about this.  I think the one concern he has about a single config file for all deployment types could be overcome.  It may require an additional tool, but that would be  a separate effort that this could easily work with.

xingzhou:
yeap, I agree. But anyway, we would still look forward to working out a separate charts deployment solution first, please let me know the developer's id so that we can have some conversation over this topic, thank you.

berndtj:
@tenczar is coming up to speed and will be working on it.

xingzhou:
@berndtj, @tenczar and @DanLavine, I created a project using Helm to deploy open whisk, now it's working on the latest open whisk build. I created a repo and uploaded the draft code, PTAL, thanks:
https://github.com/xingzhou/Deploy_OpenWhisk_with_Helm

berndtj:
Awesome @xingzhou!  I'm going to close this WIP.

lzbj:
Amazing, @xingzhou ,great job!

PR #46, opened 2017-07-27
 Deploy CouchDB manually

DanLavine:
* This gets rid of the need for the OpenWhisk configuration image 
* Update Docker images so there is no more configuration image 
* Remove specific Instructions for Minikube since everything is   deployed via yaml. The only 
  properteis that should now change is an additional environment variable in the invoker.yml 
* Update CI


kubernetes/couchdb/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+kubectl apply -f couchdb.yml
+```
+
+Once the Pod is up and running, the pod then needs to be
+configured. To do this, you need to run the configuration
+pod:
+
+```
+kubectl create -f couchdb-configure.yml
```
dgrove-oss: Minor nit:  The file is actually named couchdb-setup.yml

DanLavine:
@ddragosd  I'm not sure exactly what I want to do for that. I was thinking if we want CouchDB to be HA, then you would want a different image to perform the configuration so you only need to run it once. I would think it makes it simpler to just deploy Couch and configure it once. Rather than having to build all that logic in later.

kubernetes/couchdb/README.md
```diff
@@ -10,23 +10,33 @@ Pod. This can be done by running:
...
 ```
 
-Once the Pod is up and running, the pod then needs to be
-configured. To do this, you need to run the configuration
-pod:
+This pod goes through the process of pulling the OpenWhisk
+repo and running through some of the ansible playbooks for
+configuring CouchDB.
+
+**NOTE** the pod will say running as soon as the start command
```
ddragosd: tiny typo `stay`

DanLavine: oops

kubernetes/couchdb/docker/init.sh
```diff
@@ -18,6 +24,30 @@ pushd /openwhisk
...
+
+  if [ $TIMEOUT -eq 25 ]; then
+    echo "failed to setup CouchDB"
+    exit 1
+  fi
+
+
   # setup and initialize DB
   pushd ansible
     ansible-playbook -i environments/local setup.yml \
```
ddragosd: is this pulled from openwhisk master ?

kubernetes/couchdb/README.md
DanLavine: I'll wait for CI to make sure it passes. If it doesn't I'll update everything in 1 PR.

kubernetes/couchdb/docker/init.sh
DanLavine: yes. I wanted to make sure it always pulls the latest OW master code. That way I don't need to keep updating the image. Of course, this also means that if it fails to deploy then the image will need to be fixed as a whole

DanLavine:
my updates to travis failed. Will fix those up

kubernetes/couchdb/docker/init.sh
```diff
@@ -0,0 +1,93 @@
...
+
+
+  # if auth guest overwrite file
+  if [ -n "$AUTH_GUEST" ]; then
+    echo "$AUTH_GUEST" > /openwhisk/ansible/files/auth.guest
+  fi
+
+  # if auth whisk system overwrite file
+  if [ -n "$AUTH_WHISK_SYSTEM" ]; then
+    echo "$AUTH_WHISK_SYSTEM" > /openwhisk/ansible/files/auth.guest
```
dgrove-oss: This should be: 

echo "$AUTH_WHISK_SYSTEM" > /openwhisk/ansible/files/auth.whisk.system

PR #33, opened 2017-07-06
REVIEW: update Travis scripts to conform to other repos. and add scancode.

mrutkows:

PR #49, opened 2017-08-17
run travis files from propert directory;add scancode again.

mrutkows:


mrutkows:
These changed were originally submitted and arpproved on https://github.com/apache/incubator-openwhisk-deploy-kube/pull/33 ; however, someone updated the setup.sh and build.sh scripts in their old folder causing a conflict in git that I could not resolve because it does not like merging changes in files that another branch moved.  So we are starting "fresh" with this PR...

PR #50, opened 2017-09-08
use public images for invoker and controller

DanLavine:
This PR uses public Invoker and Controller images and it is able to create the required Kafka topics because Kafka will auto create them if they do not exist. however, it does not create the Kafka topics with all the additional parameters Ansible is able to do. For that work, we would need this [PR](https://github.com/apache/incubator-openwhisk/pull/2545) to be merged.


dgrove-oss:
lgtm;  I had done something very similar as well.  @ddragosd could you review so this can get merged?

PR #51, opened 2017-09-08
Change invoker and controller to using public openwhisk image

dgrove-oss:
Consul has been removed upstream; we can use the openwhisk images now.
Update required environment variables for controller to match requirements
of latest openwhisk image.


dgrove-oss:
My bad; didn't notice Dan beat me too it.  I need to be watching this repo more carefully!

PR #57, opened 2017-09-20
Fix Travis env variables after .travis.yml simplification

bbrowning:
The syntax here was technically incorrect but did happen to work for
some time. However, recent Travis builds have failed because
TRAVIS_KUBE_VERSION is not defined and thus `kubectl` never gets
downloaded nor Kubernetes started.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,8 @@
 sudo: required
 
 env:
-  matrix:
-    - env:
-      - TRAVIS_KUBE_VERSION=v1.6.2
-      - TRAVIS_ETCD_VERSION=v3.0.17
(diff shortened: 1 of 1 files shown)


bbrowning:
Hmm - it appears the docs at https://docs.travis-ci.com/user/environment-variables/ are wrong and this changed caused Travis to create a matrix of two builds, one with each env variable set. Odd, and not what I intended. I'm going to leave this PR open because it's the only way I know to easily get Travis to test the changes.

dgrove-oss:
can be closed; change was folded into #58

daisy-ycguo:
Closed as it has been merged with #58.

PR #62, opened 2017-09-28
refresh couchdb docker image

dgrove-oss:
Pick up the fix from #60 to init.sh and publish updated
docker image.  Switch to my dockerhub private image for
easier maintenance.


daisy-ycguo:
LGTM.

PR #64, opened 2017-09-29
Use env variable other than hard code node name

daisy-ycguo:
The node name `couchdb@couchdb0` here should use the environment variable `NODENAME` defined in kubernetes/couchdb/couchdb.yml, other than hard coded one.

Or else, it will cause the setting of `reduce_limit` fail because of `no such node: couchdb@couchdb0`.


dgrove-oss:
LGTM; tested by pulling patch into dgrove/whisk_couchdb:latest and deploying.

dgrove-oss:
@rabbah - please review & merge

PR #65, opened 2017-09-29
remove outdated TODOs

dgrove-oss:
remove TODOs that were handled by 335f20f7


dgrove-oss:
folded into #66 since they both touched controller.yaml

PR #68, opened 2017-10-12
Update counchdb image init script to avoid python package installatio…

xingzhou:
…n error.

Due to some package updates, the pip tool need to be updated before
installing ansible 2.3. Use pip install -U pip to update the package
to avoid package installation error during the start of couchdb container.

### DIFF
--- a/kubernetes/couchdb/docker/init.sh
+++ b/kubernetes/couchdb/docker/init.sh
@@ -8,6 +8,8 @@ pushd /openwhisk
   # Install ansible requirements
   ./tools/ubuntu-setup/pip.sh
 
+  sudo pip install -U pip
+
   # upgrade cffi for ansible error on Debian Jesse
   pip install --upgrade cffi
   sudo pip install markupsafe


jcrossley3:
I ran into this, too. I fixed it slightly differently: `pip install --upgrade setuptools`
I also changed the travis `build.sh` in our fork to build the image prior to applying the resource file: https://github.com/projectodd/incubator-openwhisk-deploy-kube/commit/71ad149dc547849f1f06b177671a914553b5dd9b
Might be a good idea to do that for all the required docker images, especially the ones not tied to a specific version of a dependency, e.g. the ones that checkout openwhisk master.

xingzhou:
yeap, @jcrossley3. /cc @dgrove-oss, your opinion on this?

Besides, also found a related open SSL issue during launching the latest couchdb image, can be addressed by upgrading the system pyopenssl package as:
```
sudo pip install -U pyopenssl
```
will update this patch to include this as well.

dgrove-oss:
Would the general problem be helped if we were routinely rebuilding & publishing the images?

rabbah:
@csantanapr has an apache jenkins job that is continuously publishing the core images to dockerhub. https://hub.docker.com/u/openwhisk/

dgrove-oss:
This is one of the images that has kube-specific changes to it (motivating my email to the dev list a little while ago).  Can we augment the jenkins job to also publish kube-specific images?

dgrove-oss:
#71 tries to fix this by pushing the update/install operations into Dockerfile so it happens during the image build.  I think if we combine that with getting these images published from the Jenkins job, it would be a better fix than adding updates/installs to init.sh

dgrove-oss:
@rabbah this can be closed as obsolete.  We're publishing openwhisk/kube-couchdb images on dockerhub now that contain a fix.

PR #70, opened 2017-10-13
disable use of docker-runc by setting INVOKER_USE_RUNC to false

dgrove-oss:
Set envvar to disable use of docker-runc to pause/resume containers
(will use docker pause/unpause instead).

PR #55, opened 2017-09-15
Change the invoker to DaemonSet

DanLavine:
This is a WIP for changing Invokers to a DaemonSet. It relies on this PR https://github.com/apache/incubator-openwhisk/pull/2689 to be merged first.

PR #72, opened 2017-10-18
Fix travis builds by combining PR70 and PR71

dgrove-oss:
Single PR that combines the fixes from #70 and #71 so we can get a clean Travis build in PR testing.

### DIFF
--- a/kubernetes/invoker/invoker.yml
+++ b/kubernetes/invoker/invoker.yml
@@ -55,6 +55,8 @@ spec:
             value: "latest"
           - name: "INVOKER_CONTAINER_NETWORK"
             value: "bridge"
+          - name: "INVOKER_USE_RUNC"
+            value: "false"
 
           # Properties for invoker image
           - name: "DOCKER_IMAGE_PREFIX"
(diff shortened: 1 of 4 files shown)


kubernetes/invoker/invoker.yml
```diff
@@ -70,6 +70,12 @@ spec:
           - name: "INVOKER_INSTANCES"
             value: "1"
 
+          # Invoker assigned name. Derived from hostname
+          - name: "INVOKER_NAME"
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
```
jcrossley3: Would `metadata.name` be a better fieldPath? Couldn't there be multiple invokers on a node? On minikube, for example, `spec.nodeName` returns 'localhost' where `metadata.name` returns `invoker-0`.

dgrove-oss: yep.  sorry, I grabbed this hunk from #55 without thinking about it too carefully and we  merged with minimal review to get the travis build passing again.  I'll submit a fix.

PR #73, opened 2017-10-18
use metadata.name for INVOKER_NAME for statefulsets

dgrove-oss:


dgrove-oss:
@jcrossley3 @rabbah -- address comment from review of #72

Issue #28, opened 2017-06-23
CouchDB should not need Ansible for configuration


DanLavine:
Ideally we need to be able to remove Ansible configuration that are used to setup CouchDB.

* Generate a Docker image with an already initialized database.
* The Docker image should have an init script that edits the authentication for unique credentials. Also edit the entries within the database with unique credentials

DanLavine:
Currently, CouchDB is setup by using a custom script and the public CouchDB image. The script pulls all of OpenWhisk down into the container and runs though the Ansible scripts to set things up. This works for local dev and test, but we need to host the image in a more official capacity on the OpenWhisk Docker account.

dgrove-oss:
#75 switches us to using an official openwhisk/kube-couchdb image

dgrove-oss:
#75 was merged; this issue can be closed.

PR #75, opened 2017-10-24
Use official openwhisk/kube-couchdb image

dgrove-oss:


dgrove-oss:
resolves issue #28

dgrove-oss:
@csantanapr -- switching to using the images you started publishing in #74

PR #78, opened 2017-11-02
Use zookeeper instead of redis for dynamic invoker id assignment

dgrove-oss:
Define ZooKeeper host and port instead of Redis host and port
in the environment of the invoker container (companion to OW #2916).


dgrove-oss:
@markusthoemmes, would you mind merging this as well?  Companion to https://github.com/apache/incubator-openwhisk/pull/2916

PR #82, opened 2017-11-15
add configuration of Ingress and restructure docs accordingly

dgrove-oss:
Add configuring Ingress as a separate step and include examples
of using a NodePort, a Single Service Ingress, and configurations
for IBM Cloud Lite Cluster and IBM Cloud Standard Cluster.

Extend nginx.conf to also allow connection over port 80 (needed
when TLS termination is handled by the Ingress).

Minor restructuring of top-level configuration instructions
to push down details of how to determine the API_HOST for
OpenWhisk into the new ingress/README.md file.

### DIFF
--- a/kubernetes/nginx/README.md
+++ b/kubernetes/nginx/README.md
@@ -3,12 +3,23 @@ Nginx
 
 # Deploy Nginx
 
-The Nginx Pod needs to be configured with custom certificates
-and nginx configuration file. To achieve this, nginx will need
-to create a Kube ConfigMap for the `nginx.conf` file and a
-Secrets resource with the certs.
+Depending on how you are deploying OpenWhisk, the Nginx pod
+may or may not need to support handling TLS termination
+for incoming requests. In production deployments, TLS termination
+will be handled by an Ingress placed in front of the Nginx service.
+In dev/test scenarios or when deploying on a single node cluster, it
+is likely that you will use a basic Ingress that does not handle TLS
+termination and therefore will need Nginx to handle it.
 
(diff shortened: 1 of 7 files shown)


dgrove-oss:
@bbrowning @DanLavine  does this change look ok to you?

dgrove-oss:
rebased to resolve conflict.

dgrove-oss:
rebased to latest master.  

@bbrowning @DanLavine, any comments before I ask @rabbah to review and merge?

bbrowning:
This looks good to me. We do something similar today by creating OpenShift Routes (similar to Ingress but not quite the same) to expose Nginx. We could add our definition here alongside the vanilla Kube and IBM examples, but there are numerous other config and image tweaks we need for OpenShift.

kubernetes/nginx/README.md
```diff
@@ -96,6 +107,13 @@ Secrets.
 
 # Create Nginx Docker Image
 
+We currently deploy a custom Nginx docker image that includes the
+OpenWhisk CLI and other downloadable artifacts. Once there are proper
+releases of these artifacts, we can switch to using a standard Nginx
+image and redirect to the official release archives for the artifacts
+we are currently storing in the custom docker image.  See the GitHub
+[issue](https://github.com/openwhisk/openwhisk/issues/2152).
```
rabbah: I think all the bundled artifacts are now externally released.

rabbah: Edit: nope, the CLI is still copied in instead of linking to the release.

@csantanapr @houshengbo is this planned for future?

PR #91, opened 2017-11-27
update for pureconfig (OpenWhisk PR 2576)

dgrove-oss:
Update invoker and controller deployments for
upstream change to use pureconfig and also define a new
required envvar, DB_WHISK_ACTIVATIONS_FILTER_DDOC.


dgrove-oss:
The nginx pod failed to start.  Not obvious why from travis logs.  Don't see why changing invoker/controller yml files would have any impact on nginx.   Try the travis build again?

PR #93, opened 2017-11-28
mirror PR 3010 "adapt timeouts of controller and nginx"

dgrove-oss:
mirror change to kube copy of nginix.conf from OW PR 3010.


dgrove-oss:
timeout waiting for nginx pod to deploy again.   @rabbah can you kick travis to try again.

PR #94, opened 2017-11-28
travis changes: use minikube and adjust sleep times

dgrove-oss:
1. Use minikube with --vm-driver=none instead of building hacked kubernetes
2. Upgrade kubernetes version from 1.6.2 to 1.7.4.
3. Reduce sleep interval in health check loops to cut build latency
4. Increase sleep time after creating nginx config map and secrets
   to try to reduce frequency of failures when deploying nginx.yml.


dgrove-oss:
Reduces travis times from 20-25 minutes down to 9-10 minutes.

Issue #96, opened 2017-11-29
Unable to connect to zookeeper.openwhisk:2181


lee212:
Hi,

It seems zookeeper is deployed but kafka is failing to connect.

```
$ kubectl get pods,services --all-namespaces=true -o wide
NAMESPACE   NAME                                   READY     STATUS             RESTARTS   AGE       IP           NODE
default     po/kube-apiserver-127.0.0.1            1/1       Running            0          9h        127.0.0.1    127.0.0.1
default     po/kube-controller-manager-127.0.0.1   1/1       Running            0          9h        127.0.0.1    127.0.0.1
default     po/kube-scheduler-127.0.0.1            1/1       Running            0          9h        127.0.0.1    127.0.0.1
openwhisk   po/apigateway-3311085443-mbc4c         1/1       Running            0          3h        172.17.0.4   127.0.0.1
openwhisk   po/controller-0                        0/1       CrashLoopBackOff   79         3h        172.17.0.7   127.0.0.1
openwhisk   po/controller-1                        0/1       CrashLoopBackOff   79         3h        172.17.0.8   127.0.0.1
openwhisk   po/couchdb-710020075-gfqdf             1/1       Running            0          8h        172.17.0.2   127.0.0.1
openwhisk   po/kafka-1473765933-v4p3c              0/1       CrashLoopBackOff   9          25m       172.17.0.6   127.0.0.1
openwhisk   po/redis-1106165648-zgxcs              1/1       Running            0          3h        172.17.0.3   127.0.0.1
openwhisk   po/zookeeper-1304892743-0zw8h          1/1       Running            0          29m       172.17.0.5   127.0.0.1

NAMESPACE   NAME             CLUSTER-IP      EXTERNAL-IP   PORT(S)                      AGE       SELECTOR
default     svc/kubernetes   10.254.0.1      <none>        443/TCP                      9h        <none>
openwhisk   svc/apigateway   10.254.90.227   <none>        8080/TCP,9000/TCP            3h        name=apigateway
openwhisk   svc/controller   None            <none>        8080/TCP                     3h        name=controller
openwhisk   svc/couchdb      10.254.249.61   <none>        5984/TCP                     8h        name=couchdb
openwhisk   svc/kafka        10.254.67.71    <none>        9092/TCP                     28m       name=kafka
openwhisk   svc/redis        10.254.10.167   <none>        6379/TCP                     3h        name=redis
[...]

PR #102, opened 2017-11-30
service discovery via service ennvars

dgrove-oss:
find services via service environment variables
instead of relying on kube-dns.


dgrove-oss:
closes #96

Issue #26, opened 2017-06-23
Kafka should not need Ansible for configuration


DanLavine:
Ideally we need to be able to remove Ansible configuration that are used to setup Kafka

* On the initial startup, Kafka should register the “health” topic.

dgrove-oss:
Summarizing offline discussion with @DanLavine and @rabbah on this issue.

We are running kafka with autocreation of topics enabled.  But that doesn't allow us to set the various configuration parameters for the topic that are set in the ansible scripts.  It is believed setting these appropriately is important for performance.  One possible path forward is to upgrade kafka to 0.11.0.x.  The 0.11.0 release of kafka introduced the org.apache.kafka.clients.admin.AdminClient API which would allow us to create the topics with the needed configuration parameters programmatically from the controller.  This would eliminate the need for external scripting.

dgrove-oss:
I submitted PR https://github.com/apache/incubator-openwhisk/pull/2948 to progress this.  If it is accepted, then we should be able to resolve issue with small additional effort.

dgrove-oss:
#2948 was merged.  PR https://github.com/apache/incubator-openwhisk/pull/2991 covers the health topic.  After it is merged, I believe we can use a vanilla kafka container (won't need custom image).

dgrove-oss:
#89 updates this project to take advantage of apache/incubator-openwhisk#2948.

dgrove-oss:
resolved by #130; can be closed

PR #116, opened 2017-12-08
reorg docker files to enable publishing multiple official images

dgrove-oss:
Reorganize to pull all docker subdirs out from under kubernetes
directory to a new top-level docker directory.

Tweak publish script to take the directory to build as a parameter
to enable us to build and publish multiple images (coming soon...).

PR #117, opened 2017-12-08
Switch to vanilla wurstmeister/kafka:0.11.0.1

dgrove-oss:
We no longer need a custom kafka container; all topics
are created dynamically by invoker/controller during startup!

Also upgrade from 0.10.2.1 to 0.11.0.1 to track upstream.


dgrove-oss:
Closes #26 

Intentionally not deleting kubernetes/kafka/docker directory to avoid conflict with #116.  Will delete from new location in followup PR once #116 is merged.

PR #119, opened 2017-12-08
consolidate apigateway and redis to single pod

dgrove-oss:
Redis is only needed by apigateway, so structure the deployment
of apigateway as a single pod container two containers and do
not expose redis outside of the pod.


dgrove-oss:
ready for review

Issue #24, opened 2017-06-23
Nginx should not need Ansible for configuration


DanLavine:
Ideally we need to be able to remove Ansible configuration that are used to setup Consul.

* Build Nginx with all OpenWhisk specific requirements (wsk, blackbox) pre-built into the Docker image.
* Script to helps generate certificates
* create a Kube ConfigMap and/or Secrets Resource from those certs and a static Nginx.conf file. Where this nginx.conf file is specific to an environment
* Have yaml file(s) for the Kube Deployment and Service which uses the generated ConfigMap

- [x] Nginx is configured via a yaml file with documentation
- [ ] Nginx Docker image is built via the OpenWhisk CI and is not a custom image

dgrove-oss:
See discussion in https://github.com/apache/incubator-openwhisk/issues/2152 of what can be done upstream.

dgrove-oss:
resolved by #130; can be closed

Issue #22, opened 2017-06-23
OpenWhisk should be deployed all through yaml


DanLavine:
A number of components will need to be configured so they are more "Dockerized" to remove the need for Ansible. General implementations should be able to use:
* Have Kube YAML files as the primary source of configuration
* Utilize ENVIRONMENT vars
* “bake” other configs into Docker images
* Determine if we need to do better health checking to determine if we should start. Do not fail to run a process

- [] [Nginx](https://github.com/apache/incubator-openwhisk-deploy-kube/issues/24)
- [] [Controller](https://github.com/apache/incubator-openwhisk-deploy-kube/issues/25)
- [] [Kafka](https://github.com/apache/incubator-openwhisk-deploy-kube/issues/26)
- [] [Invoker](https://github.com/apache/incubator-openwhisk-deploy-kube/issues/27)
- [] [CouchDB](https://github.com/apache/incubator-openwhisk-deploy-kube/issues/28)
- [] [RouteMgmt](https://github.com/apache/incubator-openwhisk-deploy-kube/issues/45)
- [x] Post Deploy script.

Components to be removed:
- [x] [Consul](https://github.com/apache/incubator-openwhisk-deploy-kube/issues/23)

The original proposal for this work can be found [here](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=70258567)

dgrove-oss:
resolved by #130; can be closed

PR #130, opened 2017-12-19
Remove obsolete nginx and kafka docker images

dgrove-oss:
Remove support for building these custom docker images;
we are now able to deploy using unmodified upstream images.


dgrove-oss:
Closes #22 
Closes #26 
Closes #24

PR #137, opened 2018-01-04
use configmap/secrets for CouchDB configuration

dgrove-oss:
Move DB authorization and configuration information to
configmap and secret to eliminate duplication of information
throughout the yml files.


dgrove-oss:
Closes #137

PR #138, opened 2018-01-04
controller/invoker configuration cleanup

### LINKED ISSUE #121: Define RUNTIME_MANIFESTS via configmap
We can make it easier to synchronize with the upstream project if we use a configmap for the RUNTIME_MANIFESTS instead of burying it within both controller.yml and invoker.yml.

dgrove-oss:
1. get runtimesManifest from configmap
2. specify action/trigger limits in configmap
3. put all misc. deployment-configurable parameters into configmaps
4. eliminate obsolete environment variables from invoker/controller.yml

Fixes #121

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Several requirements must be met for OpenWhisk to deploy on Kubernetes.
 ## Using Minikube
 
 For local development and testing, we recommend using Minikube version 0.23+
-with the docker network in promiscuous mode. Our Travis CI testing using Minikube 0.23.0.
+with the docker network in promiscuous mode. Our Travis CI testing uses Minikube 0.23.0.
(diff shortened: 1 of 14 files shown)


dgrove-oss:
Fixes #138 
Fixes #121

dgrove-oss:
My thinking was that the manifest was likely to get cloned & owned, since many deployments will probably want to customize their set of action runtimes.  Now that it is in a standalone file in both repositories it should be easier to keep the defaults in synch with each other.

We could probably avoid having a copy of the file here by changing the instructions to create the configmap to tell people to download the file from the main repository via a wget/curl and then using that file to create the configmap.  Have to do the same thing in build.sh for the travis builds.  Doable, but I'm not sure it is worth it.  WDYT?

PR #139, opened 2018-01-04
bump minikube version to latest release (0.24.1)

dgrove-oss:
Upgrade minikube version used for TravisCI testing
to the latest released version in hopes of reducing
the frequency of minikube crashes during CI runs.


dgrove-oss:
0.24.1 seems even less stable than 0.23.0.  Both travis CI runs failed attempting to start the first pod (couchdb).  Giving up for now.

PR #140, opened 2018-01-05
use runtimesManifest to drive invoker 'docker pull'

### LINKED ISSUE #135: invoker: get images to pull from runtimeManifests
We should be using the list of non-deprecated images in runtimeManifests to drive the docker pull operations.  Once #121 is fixed, use the same config map to drive the docker pull.  

One options is to use an auxillary init container running in the invoker pod to parse the json and do the pull operations (instead of a hardwired shell command in a postStart action).

Alternatively, we could inject a simple shell script to do the pulls into the upstream invoker container so it can be invoked here as a post start action from a single container.

dgrove-oss:
Replace hard-wired list of images to pull from invoker.yml
by an initContainer that interprets runtimesManifest to
determine which images to pull before starting the main
invoker container.

Fixes #135

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -8,8 +8,11 @@ The built images are:
   * couchdb - creates and initializes a CouchDB instance for
     dev/testing of OpenWhisk.  This image is not intended for
     production usage.
-  * routemgmt - installs OpenWhisk's route management package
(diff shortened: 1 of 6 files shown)


dgrove-oss:
@rabbah - ready for review & merge

PR #141, opened 2018-01-05
exit job's init.sh with error if subcommand fails

### LINKED ISSUE #136: error checking on install catalog/routemgmt jobs
Need to harden script so that if any of the `wsk` commands fail, the job exits with an error code.  Currently the script isn't checking the exit code of `wsk` and thus the job reports success even if the installation of packages/actions fails miserably.

dgrove-oss:
Fixes #136


dgrove-oss:
@rabbah - ready for review & merge

PR #153, opened 2018-01-16
Add support for persistent storage.

### LINKED ISSUE #39: Openwhisk fails after cluster reboot due to no persistence
I'm running a Kubernetes cluster in Vagrant for testing and noticed that on reboot, openwhisk fails because  the couchdb and consul data is not persisted.

I've already made the required changes in my fork and it's all working.
PR incoming.

dgrove-oss:
CouchDB, zookeeper, kafka, and redis all need
persistent storage to properly handle crashes
and pod restarts. Define the needed PersistentVolumes
and PersistentVolumeClaims, mount them into the
appropriate pods, and update docs accordingly.
The existing definition is appropriate for minikube
or for deploying on single-node development/test
clusters.

Production environments will need to modify
persistent-volumes.yml to meet their needs.
Also note that if multiple replicas of zookeeper,
redis, and/or kafka are deployed, then nonpersistent
storage may be used for their data volumes
(replication replaces persistence).

Fixes #39.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,12 +62,13 @@ use to deploy OpenWhisk on Kubernetes for our Travis CI testing.
 
 Do one of the following:
 * For development and testing purposes, this repo includes a configuration
-  for deploying a [simple non-persistent CouchDB instance](kubernetes/couchdb)
+  for deploying a [non-replicated CouchDB instance](kubernetes/couchdb)
   within the Kubernetes cluster.
 * For a production level CouchDB instance, take a look at the main
   OpenWhisk [documentation for configuring CouchDB](https://github.com/apache/incubator-openwhisk/blob/master/tools/db/README.md).
   You will need to define the db.auth secret and db.config configmap as described in the [CouchDB README.md](kubernetes/couchdb/README.md)
-  to match your database deployment.
+  to match your database deployment and create a CouchDB service instance
+  that forwards connections to your external database.
 
 ## Deploy Remaining Components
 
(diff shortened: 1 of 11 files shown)


dgrove-oss:
@bbrowning @jcrossley3 -- could one of you sanity check this PR?  One of my first forays into using PersistentVolumes/PersistentVolumeClaims, so a second look by someone who knows kube would be useful.

ruromero:
@dgrove-oss as you have added the PVC reference in the Deployment objects I'd add the PVC definition in the same file as each deployment.
As any user might want to use different persistent storage (e.g. NFS, iSCSI, etc.) and/or of different size. It is fine to provide an example with hostPath but as they have the same accessMode and size (almost), they will end up being bound with mismatching names. E.g. pv-zookeeper-datalog-01 with pv-apigateway-01. It can also be the case where the PVCs requesting 1Gi are bound to the 2Gi PVs leaving the other PVCs of 2Gi without any available PV.
Therefore the PV name is not meaningful unless you force it with [StorageClass](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class) or [selector.matchLabels](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims)

dgrove-oss:
@ruromero - Thanks for the review.  I think I will update to force it with selector.matchLabels.  I was hoping to minimize the number of files that the user would need to edit during deployment (which is why I put the PVCs outside of the main deployment files, which I'd like the user to be able to apply unchanged).

dgrove-oss:
@ruromero - I added the matchLabels this afternoon;  does this look ok to you now?

csantanapr:
bump @dgrove-oss

dgrove-oss:
@csantanapr -- I think this PR is ready to merge.  I rebased & squashed.  Could you review?

PR #154, opened 2018-01-23
container network/dns converted to PureConfig

dgrove-oss:
update envvar names for OpenWhisk PR#3198 which moved
network and DNS container args to PureConfig.

### DIFF
--- a/kubernetes/invoker/invoker.yml
+++ b/kubernetes/invoker/invoker.yml
@@ -93,12 +93,12 @@ spec:
                 key: api_host
 
           # Docker-related options
-          - name: "INVOKER_CONTAINER_NETWORK"
+          - name: "CONFIG_whisk_containerFactory_containerArgs_network"
             valueFrom:
               configMapKeyRef:
                 name: invoker.config
                 key: invoker_container_network
-          - name: "INVOKER_CONTAINER_DNS"
+          - name: "CONFIG_whisk_containerFactory_containerArgs_dnsServers"
             valueFrom:
               configMapKeyRef:
                 name: invoker.config


dgrove-oss:
@tysonnorris -- could you review and merge?  Thanks!

kubernetes/invoker/invoker.yml
```diff
@@ -93,12 +93,12 @@ spec:
...
+          - name: "CONFIG_whisk_containerFactory_containerArgs_network"
             valueFrom:
               configMapKeyRef:
                 name: invoker.config
                 key: invoker_container_network
-          - name: "INVOKER_CONTAINER_DNS"
+          - name: "CONFIG_whisk_containerFactory_containerArgs_dnsServers"
             valueFrom:
               configMapKeyRef:
                 name: invoker.config
```
tysonnorris: Sorry for the delay! 
I can't tell whether configMapKeyRef will 'do the right thing' here; in this case 'the right thing' may not be obvious, but should be something like: 
```
CONFIG_whisk_containerFactory_containerArgs_dnsServers_0=8.8.8.8
CONFIG_whisk_containerFactory_containerArgs_dnsServers_1=8.8.4.4
```
AFAIK this is the only way to produce an array typed value in lightbend config, and currently this is an array typed value.

dgrove-oss: ugh.  ok, didn't realize that.  No, this won't do the right thing.  Going to close this PR and open an issue to get back to it at some point later.

tysonnorris: BTW - a couple options I can think of here:
* update kube configs to explicitly provide each array entry (no change non-kube change required, but I'm not sure if this complicates things for the kube configs)
* update transformEnvironment to do the needful when a var ends with `.array`, e.g. 
```
CONFIG_whisk_containerFactory_containerArgs_dnsServers.array="8.8.8.8,8.8.4.4" -> 
CONFIG_whisk_containerFactory_containerArgs_dnsServers_0=8.8.8.8
CONFIG_whisk_containerFactory_containerArgs_dnsServers_1=8.8.4.4
```
* provide a custom parser to pureconfig https://pureconfig.github.io/docs/supporting-new-types.html . I did this originally for the array handling but since there is some basic support in lightbend config, removed it in favor of the lightbend array impl (where env vars ending with `.<index>` are treated as array type).

dgrove-oss: I like option 2 (enhance transformEnvironment to do extra processing for variables ending in .array).  @markusthoemmes, what do you think?

PR #155, opened 2018-01-26
config files to use kubernetes container pool and invoker-agent

### LINKED ISSUE #110: document architectural options for Invoker deployment
This issue is to document options for deploying the invoker subsystem for OpenWhisk.  The topic has been discussed in various venues before, most recently in a review of #107 by @stigsb. 

The key choice to make when deploying invokers is what implementation of the ContainerFactoryProvider SPI to use.   There are currently two approaches being used by downsteam consumers of this project:

## DockerContainerFactory
In this approach, the Kubernetes scheduler is only used to deploy the OpenWhisk "control plane". All of the user action containers are created, managed, and destroyed by the invoker using docker on the Kubernetes worker node.  For this approach to work well, it is essential that there is exactly 1 invoker pod per worker node that is intended for user function execution.   Using a Daemonset for the invokers is a natural fit, since the nodes intended for the invoker to use will be fairly static and can be labeled accordingly. Capacity is added/removed from the system by adding/removing worker nodes to the cluster and/or adding/removing the invoker label to the worker nodes.

This approach has the advantage of supporting low latency suspend/resume operations, but gives up some of the advantages of running on Kubernetes because it keeps the Kubernetes scheduler in the dark and forces a relatively static allocation of worker nodes to OpenWhisk invokers. 

## KubernetesContainerFactory
In this approach, the Kubernetes scheduler is used for all container operations: both control plane and user containers are created, managed, and destroyed by Kubernetes.  In this approach, it is highly likely that the number of invoker pods will be much smaller than the number of worker nodes in the cluster. Furthermore, it is likely that some form of autoscaling could be applied to dynamically vary the number of invokers to match system load (although #84 is needed to really make autoscaling work well). 

This approach allows better sharing of compute resources between OpenWhisk and other uses of the Kubernetes cluster.  However, the current KubernetesContainer (https://github.com/projectodd/incubator-openwhisk/blob/d2eb77aac212fb9970f3c9f914bf5863dcbefe50/core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala#L105 and https://github.com/projectodd/incubator-openwhisk/blob/d2eb77aac212fb9970f3c9f914bf5863dcbefe50/core/invoker/src/main/scala/whisk/core/containerpool/kubernetes/KubernetesContainer.scala#L108) does not actually implement the suspend/resume actions, so cannot be used if suspension of warm containers is a deployment requirement.

dgrove-oss:
config files for deploying an invoker that uses the k8s container pool of https://github.com/apache/incubator-openwhisk/pull/3219.  

Also adds an invoker-agent that can perform node-level operations like pause/unpause and log extraction on the worker node on behalf of a potentially remote invoker pod.

Fixes #110.


dgrove-oss:
@dubeejw -- would you have time to look at the Go piece of this PR?  This is my first Go program, so I'm looking for style feedback and someone to catch newbie mistakes.

glikson:
Would it make sense to (eventually) use CRI instead of direct interaction with the docker daemon?

glikson:
Would it make sense to have invoker-agent in a different repo? It doesn't seem to be too related to kube deployment - at least at the moment..

docker/invoker-agent/main.go
```diff
@@ -0,0 +1,329 @@
...
+
+	if timeOps {
+		end := time.Now()
+		elapsed := end.Sub(start)
+		fmt.Fprintf(os.Stdout, "Unpause took %s\n", elapsed.String())
+	}
+}
+
+// handler for /resume route
+func suspendUserAction(w http.ResponseWriter, r *http.Request) {
```
rabbah: can you add to the comment/function docs what the expected contract is? ie receive container value as a parameter.

should this have some defensive logic in case of a missing container value?

docker/invoker-agent/main.go
```diff
@@ -0,0 +1,329 @@
...
+
+	vars := mux.Vars(r)
+	container := vars["container"]
+	dummy := strings.NewReader("")
+	resp, err := client.Post("http://localhost/containers/"+container+"/pause", "text/plain", dummy)
+	if err != nil {
+		w.WriteHeader(500)
+		fmt.Fprintf(w, "Pausing %s failed with error: %v\n", container, err)
+	} else if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		w.WriteHeader(resp.StatusCode)
```
rabbah: post status of 500? i'd think the contract to the caller should be success or failure.

docker/invoker-agent/main.go
```diff
@@ -0,0 +1,329 @@
...
+		fmt.Fprintf(os.Stdout, "LogForward took %s\n", elapsed.String())
+	}
+}
+
+/*
+ * Suppout for suspend/resume operations
+ */
+
+// handler for /resume route
+func resumeUserAction(w http.ResponseWriter, r *http.Request) {
```
rabbah: similar comments to suspend below.

docker/invoker-agent/main.go
```diff
@@ -0,0 +1,329 @@
...
+	w.WriteHeader(code)
+	fmt.Fprint(w, msg)
+	fmt.Fprintln(os.Stderr, msg)
+	if metadata != "" {
+		writeSyntheticLogLine(genericLogErrorMessage, metadata)
+	}
+}
+
+// Request handler for /logs route
+func forwardLogsFromUserAction(w http.ResponseWriter, r *http.Request) {
```
rabbah: it would be nice to test this logic - perhaps a slight refactoring to avoid having to mock the http call would make a unit test easier to write.

docker/invoker-agent/main.go
```diff
@@ -0,0 +1,329 @@
...
+	var start time.Time
+	if timeOps {
+		start = time.Now()
+	}
+
+	vars := mux.Vars(r)
+	container := vars["container"]
```
rabbah: defensively error out if container is not set?

docker/invoker-agent/main.go
```diff
@@ -0,0 +1,329 @@
...
+
+	var lfi LogForwardInfo
+	b, err := ioutil.ReadAll(r.Body)
```
rabbah: what's expected in the body (looks like last seek position) - can you document in the function docs?

dgrove-oss:
added basic documentation to the entrypoint functions & opened an issue to add unit tests for invokerAgent later.

dgrove-oss:
@rabbah -- I added a little documentation on the key methods and opened an issue to document need for unit testing of log extraction of invokerAgent.  Good enough for now?

PR #184, opened 2018-04-17
helm: get runtimes from runtimes.json

dgrove-oss:
1. The set of runtimes is now read from runtimes.json
(or runtime-minimal-travis.json) instead of being hardwired.

2. Add init container to invoker_deployment to do docker pull of runtimes.

### DIFF
--- a/helm/runtimes-minimal-travis.json
+++ b/helm/runtimes-minimal-travis.json
@@ -1,0 +1,28 @@
+{
+    "runtimes": {
+        "nodejs": [
+            {
+                "kind": "nodejs:6",
+                "default": true,
+                "image": {
+                    "name": "nodejs6action"
+                },
+                "deprecated": false
+            }
+        ],
+        "python": [
+            {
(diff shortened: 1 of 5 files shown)


dgrove-oss:
@xingzhou -- does this change look sensible to you?  Seems to do what I want, but I'm still early on learning helm.

xingzhou:
@dgrove-oss, I remembered we used to use `lifecycleHook` to download the runtime once a invoker instance is started. So compared to `initContainer`, which one do you think is better?

xingzhou:
other parts looks great to me

dgrove-oss:
initContainer is closer to what the ansible-based deploy does (the invoker isn't ready until the runtimes are downloaded).   It does add some latency to starting, but avoids dispatching work to an invoker that isn't really ready to execute it yet.

thanks for looking over the changes!

xingzhou:
That makes sense, thanks david, /lgtm

PR #186, opened 2018-04-19
Helm chart improvements

dgrove-oss:
A number of fixes and improvements:
 - updates to README
 - pass through all deployments updating env vars to match master
 - define affinity and anti-affinity helper templates
 - improved readiness probes; move to template file
 - change invoker to demonset (needed to get stable invoker names)
 - fold redis container into apigateway pod
 - enable apigateway in deployment


dgrove-oss:
@csantanapr @rabbah -- would one of you be willing to review & merge if the changes look reasonable to you?

Not doing everything the non-Helm install is doing yet, but assuming the docker images are already downloaded, it gets from nothing to a working basic install on minikube in about 3 minutes with just a couple of commands...

PR #187, opened 2018-04-24
include Helm-based deployment in TravisCI matrix

dgrove-oss:


dgrove-oss:
@rabbah @csantanapr   -- could one of you please review this PR.  I finally managed to find the right set of kludges to get Helm running minikube under TravisCI on 14.04.  So it is now ready to be merged.   Thanks!

PR #193, opened 2018-05-10
Flatten and simplify Helm charts

dgrove-oss:
Consolidate into a single level of charts to simplify structure.


csantanapr:
Problems with Travis

helm/Chart.yaml
```diff
@@ -5,5 +5,7 @@ version: 0.1.0
 keywords:
   - Apache OpenWhisk
 maintainers:
+  - name: Dave Grove
```
csantanapr: We should personal names and emails from ASF repos. H
How about:
Name: Apache OpenWhisk committers
email: dev@openwhisk.org

PR #194, opened 2018-05-10
Adopt latest scancode ASF release config. (enforce *.md, .conf, .sh).

mrutkows:

PR #196, opened 2018-05-11
perform all downloads in /tmp to avoid adding spurious files into clo…

dgrove-oss:
…ned repo


dgrove-oss:
we could move scancode.sh earlier, but it is also probably better to the install of pre-req software in setup.sh in /tmp instead of inside the repo we are testing.  (see discussion in #194)

PR #210, opened 2018-06-04
Helm: Enhance pod configuration

dgrove-oss:
Take a pass through all the helm yaml files to align pod environment variables and configuration with the upstream project and to move all authorization names/passwords into Kubernetes secrets.

Enhance templates/NOTES.txt to print out the exact wsk cli command to issue to correctly set the apihost property to connect to this instance.

Split redis/apigateway into their own pods to enable separate scaling.

Add support for persistent volumes to zookeeper, kafka, couchdb, and redis pods using dynamic volume provisioning.


helm/templates/NOTES.txt
```diff
@@ -1,5 +1,9 @@
 Thank you for installing {{ .Chart.Name }}.
 
+To configure your wsk cli to connect to it, set the apihost property
+using the command below:
+  $ wsk property set --apihost {{ .Values.whisk.ingress.api_host_name -}}:{{- .Values.whisk.ingress.api_host_port }}
```
csantanapr: ❤️

PR #212, opened 2018-06-06
Reorganize docs to steer users to Helm

dgrove-oss:
Reorganize the documentation to make Helm the recommended way of
deploying OpenWhisk on Kubernetes. In the process, take an editing
pass through the main existing documentation files and try to improve
the flow and organization. Also consolidate common problems into a
troubleshooting guide.


dgrove-oss:
@rabbah @csantanapr could one of you give this a review?  Thanks!

README.md
```diff
@@ -26,23 +26,23 @@ This repository can be used to deploy OpenWhisk to a Kubernetes cluster.
...
+# Setting up Kubernetes and Helm
+
+## Kubernetes
+
+### Requirements
 
-**Kubernetes**
-* [Kubernetes](https://github.com/kubernetes/kubernetes) version 1.6+. However, multiple minor releases of Kubernetes, including 1.6.3, 1.7.14, 1.8.9 and 1.9.4 will not work for OpenWhisk due to bugs with volume mount subpaths (see[[1](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.6.md#known-issues-for-v163), [2](https://github.com/kubernetes/kubernetes/issues/61076)]. This bug will surface as a failure when deploying the nginx container.
+Several requirements must be met for OpenWhisk to deploy on Kubernetes.
+* [Kubernetes](https://github.com/kubernetes/kubernetes) version 1.7+. However, multiple minor releases of Kubernetes, including 1.7.14, 1.8.9 and 1.9.4 will not work for OpenWhisk due to bugs with volume mount subpaths (see[[1](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.6.md#known-issues-for-v163), [2](https://github.com/kubernetes/kubernetes/issues/61076)]. This bug will surface as a failure when deploying the nginx container.
```
csantanapr: We don't even test version 1.7, no way to know if it works or not.

I would suggest to put here the minimum version we know it work, still point out the problems with volume mount paths.

What about 

* [Kubernetes](https://github.com/kubernetes/kubernetes) version 1.8.10+, 1.9.5+, and 1.10.x
The following releases of Kubernetes, including 1.8.9 and 1.9.4 will not work for OpenWhisk due to bugs with volume mount subpaths (see[[1](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.6.md#known-issues-for-v163), [2](https://github.com/kubernetes/kubernetes/issues/61076)]. This bug will surface as a failure when deploying the nginx container.

README.md
```diff
@@ -56,88 +56,147 @@ Travis CI testing.
...
+Deployment can be done by using the following single command:
+```shell
+helm install . --namespace=openwhisk -f mycluster.yaml
+```
 
-## Install system actions and the openwhisk catalog
+After a while, if you can see all the pods listed by the command below
+are in `Running` state, congratulations, you have finished OpenWhisk
+deployment:
+```shell
```
csantanapr: I think is better to suggest for user to use `helm status` first, and then as secondary to use `kubectl`

README.md
```diff
@@ -56,88 +56,147 @@ Travis CI testing.
...
-pv-owner=openwhisk:
+# Cleanup
 
+Use the following command to remove the deployment:
+```shell
+helm delete <release_name>
 ```
-kubectl delete namespace openwhisk
-kubectl delete persistentvolume -lpv-owner=openwhisk
+or with a `--purge` option if you want to completely remove the deployment from helm:
```
csantanapr: maybe explain wha't the difference between the commands `remove the deployment` vs. `completely remove the deployment`

csantanapr:
@dgrove-oss feel free to merge this version, or make adjustments and merge

csantanapr:
the troubleshooting doc is very useful

README.md
dgrove-oss: Good point.  I dropped 1.7 and made some of the other adjustment.  I need to look into what minor versions of 1.8.x, 1.9.x, and 1.10.x we should recommend.

README.md
dgrove-oss: done.

README.md
dgrove-oss: good point. done.

PR #234, opened 2018-06-25
Update for upstream move of version info to PureConfig

dgrove-oss:
Update Helm chart for upstream PR 3617 that eliminated
whisk_version name and moved the version number and date
to PureConfig.


dgrove-oss:
Note.  This PR will not pass Travis until a new openwhisk/controller image is published on DockerHub

PR #236, opened 2018-06-27
Helm support for clustered zookeeper

dgrove-oss:
Allow zookeeper.replicaCount to be set to values > 1 by
properly configuring the zookeeper cluster and adjusting other
parts of the Helm chart so that dependent services can
properly communicate with the cluster.

Also document which of the deployments do not yet support
multiple replicas in docs/configurationChoices.md and
via comments in values.yaml.

### DIFF
--- a/helm/templates/zookeeper.yaml
+++ b/helm/templates/zookeeper.yaml
@@ -11,6 +11,7 @@ metadata:
 spec:
   selector:
     name: {{ .Values.zookeeper.name | quote }}
+  clusterIP: None
   ports:
     - port: {{ .Values.zookeeper.port }}
       name: "zookeeper"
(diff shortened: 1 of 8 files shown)


helm/templates/zookeeper.yaml
```diff
@@ -20,14 +21,35 @@ spec:
...
+  name: {{ .Values.zookeeper.name | quote }}
+  namespace: {{ .Release.Namespace | quote }}
+data:
+  zoo.cfg: |
+    tickTime=2000
+    clientPort={{ .Values.zookeeper.port }}
+    initLimit=5
+    syncLimit=2
+    dataDir=/data
+    dataLogDir=/datalog
```
daisy-ycguo: I wonder if these values can be moved to values.yaml. If they can, it's good to have all configuration variables in a single place.

dgrove-oss: Good point -- thanks!  Done.

helm/values.yaml
```diff
@@ -166,6 +170,8 @@ invoker:
 apigw:
   name: "apigateway"
   image: "openwhisk/apigateway:latest"
+  # sett
```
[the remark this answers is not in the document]
dgrove-oss: fixed.

dgrove-oss:
Addressed review comments, rebased, & squashed.  Should be ready to merge as soon as Travis is happy.

PR #237, opened 2018-06-28
Final removal of non-Helm based deployment

dgrove-oss:
Helm-based deployment has strictly more capabilities than the non-Helm
based configuration files.  We already stopped testing the non-Helm
deployment and it has now become out-of-date with the upstream code
base.  Therefore it is time to remove the old style files to reduce
maintenance costs and avoid user confusion.


dgrove-oss:
Rethinking.  Will do this in two PRs.  First a small one to remove the travisCI dependency on getting guest.auth and then a second pure removal of the kubernetes subtree.

PR #239, opened 2018-06-28
Final removal of non-Helm based deployment

dgrove-oss:
Helm-based deployment has strictly more capabilities than the non-Helm
based configuration files.  We already stopped testing the non-Helm
deployment and it has now become out-of-date with the upstream code
base.  Therefore it is time to remove the old style files to reduce
maintenance costs and avoid user confusion.


dgrove-oss:
Posted email to dev list letting people know about this PR.  Let's wait 48 hours before merging in case there are comments/objections.

dgrove-oss:
All comments on the dev list were in favor of the removal.  @rabbah - could you go ahead and approve this PR?

PR #241, opened 2018-07-02
store key OpenWhisk configuration values in ConfigMaps

dgrove-oss:
Go back to using ConfigMaps to store key cluster and database
configuration parameters and modify deployments to access the data
from those maps. Having explicit config maps makes it easier to
inspect the cluster configuration and enable separating the core
system and the providers into separate charts (subsequent PR after
this is merged).

Note: database and system authentication information continues to
be stored in secrets; that is unchanged by this PR.

### DIFF
--- a/helm/templates/_helpers.tpl
+++ b/helm/templates/_helpers.tpl
@@ -53,17 +53,35 @@
       name: db.auth
       key: db_password
 - name: "CONFIG_whisk_couchdb_port"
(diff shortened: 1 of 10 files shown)


daisy-ycguo:
@dgrove-oss I think we should also update document to affect the changes. We should document how to set up these configure maps. Do you want to use a separate PR to update document, or you want to include in this PR ?

PR #262, opened 2018-07-19
add check for healthy invoker to build-helm.sh

dgrove-oss:
Use the controller's /invokers/healthy/count route to verify
that the controller has at least one healthy invoker before
attempting to proceed to the smoketest of invoking an action.


dgrove-oss:
investigating travisci failure.  initial guess is that kube-deploy got out-of-synch with upstream repo while I was on vacation...

dgrove-oss:
rebased to pick up fixes from #268

Issue #267, opened 2018-07-31
Invoker failling with error


cduque89:
After fixing #266 we started to have the following error on the invoker:

````
Picked up JAVA_TOOL_OPTIONS: -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap
[2018-07-31T12:56:30.381Z] [INFO] Slf4jLogger started
[2018-07-31T12:56:30.999Z] [INFO] [#tid_sid_unknown] [Config] environment set value for whisk.api.host.name
[2018-07-31T12:56:31.003Z] [INFO] [#tid_sid_unknown] [Config] environment set value for whisk.api.host.proto
[2018-07-31T12:56:31.005Z] [INFO] [#tid_sid_unknown] [Config] environment set value for whisk.api.host.port
[2018-07-31T12:56:31.006Z] [INFO] [#tid_sid_unknown] [Config] environment set value for zookeeper.hosts
[2018-07-31T12:56:31.006Z] [INFO] [#tid_sid_unknown] [Config] environment set value for runtimes.manifest
[2018-07-31T12:56:31.007Z] [INFO] [#tid_sid_unknown] [Config] environment set value for kafka.hosts
[2018-07-31T12:56:31.008Z] [INFO] [#tid_sid_unknown] [Config] environment set value for port
[2018-07-31T12:56:31.176Z] [INFO] [#tid_sid_unknown] [Invoker] Command line arguments parsed to yield CmdLineArgs(None,None,None)
[2018-07-31T12:56:31.179Z] [ERROR] [#tid_sid_invoker] [Invoker] Either --id or --uniqueName must be configured with correct values
````
It seems a parameter is missing somewhere, can anyone help?

Thanks.

markusthoemmes:
Argh, sorry forgot to update.

We basically need to replay this commit: https://github.com/apache/incubator-openwhisk/commit/b197332e041968e98f6cec996e4a136116a90569 to the kubernetes repo. Want to open a PR on that? :)

cduque89:
Hi,
Thanks. Will try it now and make a PR.

Thanks.

cduque89:
Hi @markusthoemmes.

We tried but now we get the following error:

````
[ERROR] [#tid_sid_invoker] [Invoker] Error processing command line arguments List(--id, 554q8, --uniqueName, 554q8)
`````

We changed the line on invoker.yaml to:
`````
[...]

Issue #266, opened 2018-07-31
Controller waiting for couchdb giving unauthorised


cduque89:
Hi.

With the latest chart version my controller readiness probe is failing because it's giving unauthorised when trying to reach couchdb.

We added the following to the readiness probe and now it works: 
````
--header=\"Authorization: Basic {{ include "db_authentication" . | b64enc }}\"
```` 
and in _helpers.tpl:
````
{{- define "db_authentication" -}}
{{ .Values.db.auth.username }}:{{ .Values.db.auth.password }}
{{- end -}}
````

Is this happening to someone else as well?

Should we open a PR?

Thanks.

dgrove-oss:
There have been a couple of upstream PRs merged that need adjustment for in kube-deploy.  i'm working on fixing now.  Should have fixes ready today.

cduque89:
We have opened a PR that fix this. #268 it fixed everything for us. If you have a look and agree with the changes you can merge it.

PR #268, opened 2018-07-31
Fix for #266 and #267

cduque89:
fixed readiness probe for couch db issue: #266
fixed invoker init.sh parameter needs uniqueName issue: #267


bbrowning:
Thanks for the fix! I just ran into these issues locally and was able to quickly verify that this fix works.

dgrove-oss:
Agreed.  @bbrowning I can do a squash and merge from the Github UI if we want this in quickly.

@cduque89 - we tend to use git rebasing instead of merge commits to make the history cleaner.  I think there are some docs floating around about the preferred workflow.  I can find them for you if that is helpful.

PR #271, opened 2018-08-09
Correctly writing custom OpenWhisk authentication keys (for the guest…

jvdgeest:
… and whisk.system users) into the CouchDB database.

With this fix, custom auth keys that are defined in `values.yml` of the Helm chart (`whisk.auth.system` and `whisk.auth.guest`) are correctly applied and stored in the CouchDB database of OpenWhisk.

In the `initdb.sh` script, the `https://github.com/apache/incubator-openwhisk` git repository is cloned on line number 5. In this git repository, the auth keys for OpenWhisk are stored in the `ansible/files/auth.guest` and `ansible/files/auth.whisk.system` files. These files contain the default authentication keys.

I have added two lines to this shell script, that take the secret keys from Kubernetes containing the custom values defined in`values.yml` of the Helm chart.

### DIFF
--- a/helm/openwhisk/configMapFiles/initCouchDB/initdb.sh
+++ b/helm/openwhisk/configMapFiles/initCouchDB/initdb.sh
@@ -4,6 +4,10 @@
 # Clone OpenWhisk to get the ansible playbooks needed to initialize CouchDB
 git clone https://github.com/apache/incubator-openwhisk /openwhisk
 
+# overwriting auth.guest and auth.whisk.system files with custom defined keys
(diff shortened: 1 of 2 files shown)


dgrove-oss:
Thanks; change makes sense to me.

Could you please file an Apache CLA as described here: https://github.com/apache/incubator-openwhisk/blob/master/CONTRIBUTING.md

dgrove-oss:
I did a slightly different version of this fix in #273 that mounts the secrets in a volume instead of putting them in envvars.

dgrove-oss:
Closing in favor of #273 -- thank you for the problem report & suggested fix!

PR #284, opened 2018-08-23
Enable cloudant package provider to use a different couchdb instance

### LINKED ISSUE #283: Enable providers to use another database other than OpenWhisk system database
We should enable a choice for providers to use another database, other than the same database used by OpenWhisk.

Charters and documents need be be updated.

daisy-ycguo:
Partially closes #283


daisy-ycguo:
Kafka provider and alarm provider can make the same change, which will be submitted in the follow up PRs.

Issue #247, opened 2018-07-05
TravisCI testing of providers


dgrove-oss:
We should add a TravisCI build stage to deploy and test the providers.  Script should run after the main build-helm.sh concludes successfully.

PR #296, opened 2018-09-13
Add travis test for providers

daisy-ycguo:
Closes Issue #247 

Including:
- Add travis test for Kafka, Alarm and Cloudant providers
- Fix small errors of providers' README doc

### DIFF
--- a/helm/openwhisk-providers/charts/ow-alarm/README.md
+++ b/helm/openwhisk-providers/charts/ow-alarm/README.md
@@ -65,7 +65,7 @@ invoker:
 
 You may install this chart with command like
 ```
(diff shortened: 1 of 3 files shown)


daisy-ycguo:
This PR doesn't include the scripts to verify triggers' creation. I don't include this because the success of trigger creation depends on too much environment settings
- Correct parameters of a `real` event source
- OW_CONTAINER_FACTORY is set to `kubernetes` or Kube DNS is configured.
- PV defined

I think it's much complicated and will affect the base OpenWhisk environment. If you think we should include that, I propose to open a new issue and create a separate PR.

dgrove-oss:
Hi Daisy, that makes sense to me.  Checking that the Helm chart deploys correctly is a good enough smoketest.  I agree actually firing a trigger is more complicated.

PR #297, opened 2018-09-14
Add optional Fluent Bit log enrichment agent

dgrove-oss:
Companion to OW core PR https://github.com/apache/incubator-openwhisk/pull/3974.

Adds an optional DaemonSet of Fuent Bit log processing agents to enrich the user container logs with activationIds.

### DIFF
--- a/helm/openwhisk/templates/_invokerHelpers.tpl
+++ b/helm/openwhisk/templates/_invokerHelpers.tpl
@@ -25,6 +25,7 @@
   mountPath: "/var/run/docker.sock"
 - name: dockerrootdir
   mountPath: "/containers"
+  readOnly: true
 {{- end -}}
 
 {{- define "docker_pull_runtimes" -}}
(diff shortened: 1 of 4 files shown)


daisy-ycguo:
I have two questions to this PR:
- Will you use a formal repo to store the image: `dgrove/ow-fluent-bit`?
- Is it possible to input providers' logs ? Some of providers' logs are stored in PVC.

dgrove-oss:
1. Once the upstream PR is merged, the image will be available as `openwhisk/ow-fluent-bit` and I'll change this PR to use the official image.
2. My guess is that the provider logs should be handled the same way as controller/invoker/kafka/etc. logs? (they are system logs, not user logs??).   We could certainly use fluent-bit to pull them out of the PVCs if we need to, but maybe the way to do that is to add a fluent-bit container to the pod with the provider container that would then forward them to the platform logging service (potentially just by writing them to its stdout??)

dgrove-oss:
Abandoning this approach.

PR #300, opened 2018-09-19
Disable some provider tests in travis if no pv is available

daisy-ycguo:
check if persistant volume is available before alarm provider and cloudant provider tests;
exist 0 if pv is not available and continue those provider tests if pv is available.


daisy-ycguo:
Now travis test lasts for 30 minutes because pv is not defined.
It waits a long time. At the end, alarm provider and cloudant provider are still `pending`.
This PR is to solve this issue.

dgrove-oss:
Closing in favor of #301

PR #304, opened 2018-09-28
Document using the Kubernetes support available in Docker 18.06

dgrove-oss:

PR #307, opened 2018-10-03
Shuffle README to highlight Kubernetes on Docker option

dgrove-oss:


dgrove-oss:
Incorporated improvements @csantanapr suggested in #304 and added a note for how to mixin locally built docker images for dev/testing.

PR #316, opened 2018-10-18
Add an option include OpenWhisk system tests in TravisCI testing  

dgrove-oss:
Unfortunately this additional level of testing is still disabled by default because although it passes fairly reliably when run on a development machine, it fails almost 100% of the time in the TravisCI environment.
    
Also include a retry loop around the `wsk action list` smoketest in buildHelm.sh and the `wsk package list` checks because there is annoyingly often a delay between the entity being created successfully and it showing up in `list` operations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,21 @@
 # Licensed to the Apache Software Foundation (ASF) under one or more contributor
 # license agreements; and to You under the Apache License, Version 2.0.
-sudo: required
+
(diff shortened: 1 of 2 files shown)


dgrove-oss:
This has reached a useful checkpoint, so I propose merging the capability to run the systemTests, but leave it disabled for now until we can investigate why the tests are so prone to failure in the TravisCI environment.

PR #318, opened 2018-10-19
proper configuration of couchdb as a single node cluster

dgrove-oss:


dgrove-oss:
This is ready to merge.   I spent some time looking into the intermittent test failures for kube-deploy in our TravisCI runs and many of them bottom out in 500 errors coming from the connection between the controller and couchdb.  I speculate that improper configuration of CouchDB might be at least partially to blame (it defaults to expecting a 3 node cluster, but we only run 1 couchdb pod).

PR #325, opened 2018-10-31
variables in values.yaml should be in camelCase

dgrove-oss:
Per the Helm best practices document [1], variables in values.yaml
should be in camelCase

[1] https://github.com/helm/helm/blob/master/docs/chart_best_practices/values.md


csantanapr:
You don’t want to know what I think about camel_case

dgrove-oss:
Not saying I like cAmELcAsE, but this is a convention rule that is enforced in some extended Helm linters, so I give up...

csantanapr:
Hold my 🍺 let me change Helm 🎃

PR #332, opened 2018-11-02
Testing Improvements

### LINKED ISSUE #127: Run OpenWhisk test suite against kube-based deployment
At first glance, I think we need to run the test suite as a Job within the cluster because some of the tests attempt to reach internal endpoints that intentionally aren't exposed outside the cluster.

So, it might look something like a Job that does the following:
1.  clone the openwhisk repo
2. generate a whisk.properties describing the deployment (use service names for hosts)
3. compile and run the tests.

dgrove-oss:
* Support for running :tests:testSystemBasic as part of `helm test`.
    
    * Enable running tests:testSystemBasic for DockerContainerFactory
      configurations (ping test still fails with KubernetesContainerFactory
      because need to fix upstream to drop CAP_NET_ADMIN on user action pod).
    
    * Change error handling strategy in myTask.sh to being mostly
      explicit to get better control and allow retries of commands
      that might be vulnerable to CouchDB's eventual consistiency.
    
    * Remove obsolete script for starting minikube in TravisCI

    * Push openwhisk core git tag forward 2 commits to pick up fix
      for org.apache package rename in tests:testSystemBasic.

    * Separate chart deployment and testing into separate scripts
      because we need to use travis_wait when running helm tests,
      but we want to be able to eagerly see the output of the deploy
      step to enable early aborts of the test run when it gets hung.

Fixes #127.


dgrove-oss:
This is finally working reliably.....I've caught my 🐋 
Doubles the wall clock time for testing a PR to about 30 minutes, but running the system tests against the kube deploy is worth adding the time.

PR #334, opened 2018-11-05
Add php72 runtime

csantanapr:


csantanapr:
@dgrove-oss Why do we have `"sentinelledLogs": false,` for java? I don't see this in the core repo for runtimes.json

dgrove-oss:
Was removed in main repo by https://github.com/apache/incubator-openwhisk/pull/3860.  Shouldn't be here either.  (Java has sentinels now).

csantanapr:
ok I will remote now as part of these clean up 👍

PR #340, opened 2018-11-06
Conform to Helm best practices for structure of templates/

dgrove-oss:
Specifically make changes to comply with the rules:
  1. Each resource definition should be in its own template file.
  2. Template file names should use dashed notation, not camelcase.
  3. Template file names should reflect the resource kind in the name.

See guidelines at https://github.com/helm/helm/blob/master/docs/chart_best_practices/templates.md.


dgrove-oss:
A largish diff, but no actual content changes (just broke up existing files into multiple files and changed names).

PLoecker:
I believe there is a bug. When you enable persistence, helm cannot parse the yaml file redis-pod.

Error message:
Error: YAML parse error on openwhisk/templates/redis-pod.yaml: error converting YAML to JSON: yaml: line 74: did not find expected '-' indicator

PR #341, opened 2018-11-07
Fix YAML parse error in redis-pod.yaml

PLoecker:
A space was removed in commit #340. That causes a parse error when persistence is enabled.


dgrove-oss:
Thanks @PLoecker!

PR #350, opened 2018-11-12
Helm/Tiller 2.9 is acceptable

dgrove-oss:


dgrove-oss:
conflicts with #352  -- will change tiller version as part of that PR to simplify merging.,

PR #351, opened 2018-11-12
Move smoketest into chart so it can be used for `helm test`

dgrove-oss:
Pull the logic for the simple hello action smoketest into the
Helm chart for OpenWhisk so it can be used to verify that the
Chart deployed correctly.


csantanapr:
I was expecting a Job not a Pod for the tests.
I'm guessing that is any pod defined in any yaml under `/templates/tests/*` ?

I was not able to find docs about this helm test pods https://docs.helm.sh/helm/#helm-test

csantanapr:
nvm
I found that is based on 
```
 annotations:
    "helm.sh/hook": test-success
```
From https://github.com/helm/helm/blob/master/docs/chart_tests.md

PR #352, opened 2018-11-12
Streamline Helm install process + linting fixes 

dgrove-oss:
1. Consolidate core and providers into a single chart.
   Structurally, the provider charts were not independent
   of the core chart so it is more in keeping with
   best practices to have a single Chart that can
   deploy both the core and the optionally enabled event
   providers.

2. Rework the TravisCI testing so that the tests are now
   available as part of the Helm chart. This enables them
   to be easily run by a user who has installed the chart
   via `helm test <release-name>` to verify their deployment was
   successful.

3. Install the catalog, routemgmt, and provider packages
   in a single Job.

4. Per Helm best practices, resources defined in the Chart
   should not explicitly define their namespace.

5. Helm will automatically create the namespace when it
   deploys the Chart, so we can remove that step from
   our README.md

6. Change minimum Tiller version from 2.10 back to 2.9

7. Uniform labeling of all resources created by the Chart.

8. A few misc linting fixes


csantanapr:
hmmm @dgrove-oss 
How easy is to use a boolean in to enable or disable a specific provider now?
Can we have the default be to not deploy any providers still? (ie. only compute)

dgrove-oss:
@csantanapr -- yes, we can enable/disable providers by flipping a single boolean per provider.  By default, the providers are all disabled in values.yaml so what gets deployed is just the core compute.

csantanapr:
Thank you @dgrove-oss

dgrove-oss:
This is now ready for review and merge.

PR #365, opened 2018-11-19
Add nodejs:10 to default set of runtimes for k8s

csantanapr:
Related to https://github.com/apache/incubator-openwhisk/issues/4123

### DIFF
--- a/helm/openwhisk/runtimes.json
+++ b/helm/openwhisk/runtimes.json
@@ -41,7 +41,7 @@
                 "image": {
                     "prefix": "openwhisk",
                     "name": "action-nodejs-v10",
-                    "tag": "d967ebe"
+                    "tag": "586f84e"
                 },
                 "deprecated": false,
                 "attached": {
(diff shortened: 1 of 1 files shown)


helm/openwhisk/runtimes.json
```diff
@@ -34,6 +34,20 @@
...
                     "attachmentType": "text/plain"
                 }
+            },
+            {
+                "kind": "nodejs:10",
+                "default": false,
+                "image": {
+                    "prefix": "openwhisk",
+                    "name": "action-nodejs-v10",
+                    "tag": "d967ebe"
```
dgrove-oss: This tag doesn't exist in dockerhub for action-nodejs-v10.  Current available tags are `e892cd8` or `586f84e`.

csantanapr: it was a copy paste error from nodejs8, Are we pinning now?
You don't like to leave on the bleeding edge?

dgrove-oss: Thanks.  Yeah, the chart is much more useful if the deployments using it are repeatable.  So I've  switched to pinning all docker tags & git clones, with a plan of updating with known good versions on a frequent basis (roughly weekly...).

PR #370, opened 2018-11-26
Fix bug in propagation of invoker.containerFactory.kubeDNS

dgrove-oss:
Bug fix to get providers working with DockerContainerFactory
using kube-dns to allow provider actions to connect to the
couchdb service running within the Kubernetes cluster.
The propagation of the kubeDNS setting wasn't happening
because of a mismatch in names in the Helm chart.

Bug fix for providers with DockerContainerFactory


helm/openwhisk/values.yaml
```diff
@@ -205,7 +205,7 @@ invoker:
         enabled: false
         port: 3233
   containerNetwork: "bridge"
-  kubeDNS: nil
+  kubeDNS: ""
```
daisy-ycguo: Here I don't see `kubeDNS` is under `containerFactory`. What I see here is `kubeDNS` is in the same level as `containerFactory`. I don't understand why you use `.Values.invoker.containerFactory.kubeDNS` in `invoker-pod.yaml`.

PR #372, opened 2018-11-27
Helm chart providers

neeraj-laad:
- [x] Fixed some issue with helm chart values.yaml
- [x] Enabled alarm, cloudant and kafka providers by default


helm/openwhisk/values.yaml
```diff
@@ -197,7 +209,7 @@ invoker:
   containerFactory:
     dind: false
     useRunc: false
-    impl: "docker"
+    impl: "kubernetes"
```
dgrove-oss: Let's leave the default here as docker due to poor cold start and log extraction performance of KubernetesContainerFactory. Change default downstream instead, or wait for #370 to get merged before enabling providers?

helm/openwhisk/values.yaml
```diff
@@ -242,14 +254,14 @@ providers:
...
-    # password: "secret"
+    host: "10.10.10.10"
+    port: 5984
+    protocol: "http"
+    username: "admin"
+    password: "secret"
   # Alarm provider configurations
   alarm:
-    enabled: false
+    enabled: true
```
dgrove-oss: @csantanapr has mentioned in the past (#352) that is preferable to not enable providers by default here.  Change default in downstream project instead?

helm/openwhisk/values.yaml
neeraj-laad: Ok.. switching it back to docker

helm/openwhisk/values.yaml
neeraj-laad: @csantanapr @dgrove-oss Is there a reason to have the providers switched off by default? I thought it would be much easier for users to get started if things worked out of the box.

neeraj-laad: Given we are switching back to docker, and we do not have an automated way to enable them using docker container factory. I'll disable them for now.. We can review this again once #370 is merged in.

Issue #374, opened 2018-11-29
Error during "Deploy with Helm"


inaspic:
I'm following the steps in the README to deploy OpenWhisk. I ran the following:

`helm install ./helm/openwhisk --namespace=openwhisk --name=owdev -f mycluster.yaml`

but I get the following error in the controller:

`Liveness probe failed: Get http://10.1.0.170:8080/ping: dial tcp 10.1.0.170:8080: connect: connection refused Back-off restarting failed container`

Here is the log:

Picked up JAVA_TOOL_OPTIONS: -XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap
[2018-11-29T01:43:35.126Z] [INFO] Initializing Kamon...
[INFO] [11/29/2018 01:43:38.471] [main] [StatsDExtension(akka://kamon)] Starting the Kamon(StatsD) extension
[2018-11-29T01:43:38.824Z] [INFO] Slf4jLogger started
[2018-11-29T01:43:39.863Z] [INFO] [#tid_sid_unknown] [Config] environment set value for limits.triggers.fires.perMinute
[2018-11-29T01:43:39.870Z] [INFO] [#tid_sid_unknown] [Config] environment set value for limits.actions.sequence.maxLength
[2018-11-29T01:43:39.871Z] [INFO] [#tid_sid_unknown] [Config] environment set value for limits.actions.invokes.concurrent
[2018-11-29T01:43:39.873Z] [INFO] [#tid_sid_unknown] [Config] environment set value for controller.instances
[2018-11-29T01:43:39.874Z] [INFO] [#tid_sid_unknown] [Config] environment set value for limits.actions.invokes.perMinute
[2018-11-29T01:43:39.875Z] [INFO] [#tid_sid_unknown] [Config] environment set value for runtimes.manifest
[2018-11-29T01:43:39.876Z] [INFO] [#tid_sid_unknown] [Config] environment set value for kafka.hosts
[2018-11-29T01:43:39.878Z] [INFO] [#tid_sid_unknown] [Config] environment set value for port
[2018-11-29T01:43:41.958Z] [INFO] [#tid_sid_unknown] [KafkaMessagingProvider] topic completed0 already existed
[2018-11-29T01:43:42.311Z] [INFO] [#tid_sid_unknown] [KafkaMessagingProvider] topic health already existed
[...]

PR #375, opened 2018-11-29
Document pv configuration in k8s-docker-for-mac

daisy-ycguo:
Resolves issue #374

### DIFF
--- a/docs/k8s-docker-for-mac.md
+++ b/docs/k8s-docker-for-mac.md
@@ -64,6 +64,21 @@ nginx:
   httpsNodePort: 31001
 ```
 
+Since OpenWhisk deployment enables PersistentVolumes by default, you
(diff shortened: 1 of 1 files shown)


dgrove-oss:
I don't believe this change is correct.  I am using docker-for-desktop routinely with dynamic provisioning of volumes and I don't need to do either of these things.

dgrove-oss:
but, i am running docker 2.0.0.0-mac78.  So perhaps it is a new feature that isn't present in 18.06.  Daisy, could you confirm what version of docker you have on your machine?

dgrove-oss:
limited version to 18.06  in #377

daisy-ycguo:
I'm using version 18.06.1-ce-mac73 (26764), a little bit old version.

PR #377, opened 2018-11-29
Docker for Mac 2.0 does not need additional configuration for PVs

dgrove-oss:


daisy-ycguo:
Thanks David. You are right. I'm using 18.06, a little bit version.

PR #378, opened 2018-11-29
cleanup specification of storageclasses and persistence

### LINKED ISSUE #376: Review the logic of DefaultStorageClass and useDynamicProvisioning
While I tried to deploy OpenWhisk with Docker in Mac, I found the dynamic provisioning didn't work by default. In [k8s doc](https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/), it is said "When a default StorageClass exists in a cluster and a user creates a PersistentVolumeClaim with storageClassName **unspecified**, the DefaultStorageClass admission controller automatically adds the storageClassName field pointing to the default storage class."

It looks like leaving `storageClassName` as blank is not **unspecified**. So I rewrite the logic of DefaultStorageClass and useDynamicProvisioning, change the pvc specification to:
```
spec:
{{- if **not** .Values.k8s.persistence.useDynamicProvisioning }}
  storageClassName: {{ .Values.k8s.persistence.defaultStorageClass }}
{{- end }}
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: {{ .Values.db.persistence.size }}
```
It works well with Docker in Mac environment.

The new logic is:
1. If dynamic provisioning is supported by the k8s cluster ( checking if there is default StorageClass by `kubectl get StorageClass` ), set `.Values.k8s.persistence.useDynamicProvisioning` to true. Without any additional configuration, the dynamic provisioning works for every pv claim.
2. If dynamic provisioning is **not** supported by the k8s cluster, set `.Values.k8s.persistence.useDynamicProvisioning` to false and set the value of `.Values.k8s.persistence.defaultStorageClass`. Manually create 5 and more pv, and then deploy OpenWhisk.

I think we should test the new logic in another k8s cluster.

dgrove-oss:
Rename persistence variables in values.yaml to better reflect
what they actually mean. Update docs to match.

Fixes #376.

### DIFF
--- a/docs/k8s-docker-for-mac.md
+++ b/docs/k8s-docker-for-mac.md
@@ -64,29 +64,6 @@ nginx:
   httpsNodePort: 31001
 ```
 
-#### Additional Configuration for Docker for Mac 18.06
-
-If you are using Docker for Mac 18.06, you will need
-to add an additional stanza to your mycluster.yaml because
-this version does not have out-of-the-box support for automatic
-dynamic provisioning of persistent volumes. This additional
-configuration is not needed if you are using Docker for Mac 2.0.0
-or later.
-
-For 18.06 you should either set the defaultStorageClass
-by adding the stanza below to mycluster.yaml:
-```yaml
-k8s:
-  persistence:
-    defaultStorageClass: hostpath
-```
-or completely disable the use of persistence by adding the stanza below:
-```yaml
-k8s:
-  persistence:
-    enabled: false
-```
-
 ## Hints and Tips
 
 One nice feature of using Kubernetes in Docker, is that the
--- a/docs/k8s-ibm-public.md
+++ b/docs/k8s-ibm-public.md
@@ -97,12 +97,17 @@ whisk:
 
 k8s:
   persistence:
-    defaultStorageClass: default
+    hasDefaultStorageClass: false
+    explicitStorageClass: default
 ```
 
 IKS does not provide a properly configured DefaultStorageClass,
 instead you need to tell the Helm chart to use the `default`
-StorageClassName as shown above.
+StorageClassName as shown above. This StorageClass does have
+a dynamic provisioner, so it is not necessary to manually create
+the PersistentVolumes. Note that it is not unusual for it to take
+several minutes for your PersistentVolumes to be created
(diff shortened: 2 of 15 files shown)


dgrove-oss:
This works for me on docker-for-mac 2.0.0 and on an IKS (1.11) cluster.

daisy-ycguo:
This works for me on docker-for-mac 18.06, yet it doesn't work for IKS (1.10.8).

docs/k8s-docker-for-mac.md
```diff
@@ -78,7 +78,8 @@ by adding the stanza below to mycluster.yaml:
 ```yaml
 k8s:
   persistence:
-    defaultStorageClass: hostpath
+    hasDefaultStorageClass: false
+    explicitStorageClass: hostpath
```
daisy-ycguo: With your PR, this stanza is not required even for docker-for-mac 18.06. I think we can delete it.

docs/k8s-ibm-public.md
```diff
@@ -97,12 +97,15 @@ whisk:
 
 k8s:
   persistence:
-    defaultStorageClass: default
+    hasDefaultStorageClass: false
+    explicitStorageClass: default
 ```
```
daisy-ycguo: this doesn't work for IKS 10.08. I don't understand why. K8s 10.08 is also a stable version. Do you think we want to support 10.08?

daisy-ycguo: I believe IKS v10.08 does support dynamic provisioning. It may need additional configuration, which is hidden somewhere. Yet I cannot figure out.

daisy-ycguo:
Sorry for my misleading messages. Finally I verified this PR worked well for IKS 10.8, just **with more patience**. The pvc binding may take 1-2 minutes. If it shows pending, it doesn't mean it fails. Just waiting for 2 minutes, it will work fine. 

Thanks, David. After you remove document about the additional configuration for Docker for Mac, I'm OK to merge this PR.

dgrove-oss:
i'm going to merge #380 first, then reconcile the changed variable names and enhance the docs as discussed with daisy.  Please hold off merging until I remove the wip label from this PR.

dgrove-oss:
ready to be merged.

PR #391, opened 2018-12-12
synchronize with upstream configuration files

dgrove-oss:
1. Remove nginx adminportal port (OW 4168).
2. Update controller/invoker limit configuration
3. Enable kafka topic configuration in values.yaml


daisy-ycguo:
@dgrove-oss I merged this PR. I think we'd better add documentation about limitations and kafka configurations.

PR #392, opened 2018-12-13
DockerContainerFactory DNS configuration

### LINKED ISSUE #382: Improve configuration of invoker.containerFactory.nameservers
When the providers are enabled with the DockerContainerFactory, we should probably default the nameserver configuration parameters of the invoker to use Kuberetes dns service.  This allows user action containers (including those used by the provider packages) to access other kubernetes services using cluster-internal dns.  The provider's actions need this to be able to reach couchdb and operate properly. 

This is a less secure setup (it allows user actions to be aware they are running within a Kube cluster and to interact with other pods running in the cluster), but is probably what a user is expecting when they deploy OpenWhisk on their own Kubernetes cluster.

Sketch of implementation, if the user didn't explicitly set invoker.containerFactory.nameservers when they deployed the chart, then we should run a script as part of invoker startup that parses /etc/resolv.conf to get the values set for the invoker pod by Kube.  We then pass those through the appropriate PureConfig env variables so that things are properly configured without a manual user step of figuring out the ClusterIP of kube-dns / core-dns service.

dgrove-oss:
By default, automatically configure the Invoker so that user action containers created by the DockerContainerFactory will have the same DNS configuration as the Invoker itself.
    
This (a) enables all event providers to be enabled by default and (b) provides a more Kubernetes-native experience since the user action containers can access other Kubernetes services
running on the cluster via the usual Kubernetes DNS mechanisms.

Fixes #382

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,16 +142,6 @@ deploy.
 4. [Configure the `wsk` CLI](#configure-the-wsk-cli). You need to
 tell the `wsk` CLI how to connect to your OpenWhisk deployment.
 
-Although you are deploying OpenWhisk to Kubernetes, by default the
-executing actions will not be aware they are running on a Kubernetes
-cluster. In particular, they will not be configured to do DNS resolution
-of Kubernetes services.  However, if you want executing actions to
-easily access services in a Kubernetes-native way, you can configure
-your OpenWhisk deployment to enable that by either using the
-[KubernetesContainerFactory](docs/configurationChoices.md#invoker-container-factory)
-or setting the value of `invoker.DNS` when you create the `mycluster.yaml`
-to customize your deployment ([see DNS options](docs/configurationChoices.md#user-action-container-dns)).
-
 ## Initial setup
 
 Indicate the Kubernetes worker nodes that should be used to execute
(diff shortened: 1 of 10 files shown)


dgrove-oss:
Can be merged independently, but not actually useful until https://github.com/apache/incubator-openwhisk/pull/4176 is merged upstream and we do a version bump of the invoker imageTag to get the change.

dgrove-oss:
added WIP tag -- upstream PR is merged, but want to rethink the approach a little before merging this.

dgrove-oss:
Since the DNS configuration support was merged upstream, we can exploit it to make our defaults more "kube native", even with the DockerContainerFactory.  The event providers can be enabled by default and user action containers can invoke other Kubernetes services on the cluster.

sciabarracom:
I am going to test the change on my deployment

sciabarracom:
I deployed in my kubernetes cluster, it deploys but when I try to set the auth I have this problem:

```
$ AUTH=$(kubectl exec -ti openwhisk-wskadmin -- wskadmin user list guest | awk '{print $1}')
$ wsk property set --auth "$AUTH" --namespace guest --apihost https://sciabarra.net:443
error: Authenticated user does not have namespace 'guest'; set command failed: Unable to create HTTP request for GET: The Cert file is not configured. Please configure the missing Cert file.
$ wsk -i property set --auth "$AUTH" --namespace guest --apihost https://sciabarra.net:443
wsk -i property set --auth "$AUTH" --namespace guest --apihost https://sciabarra.net:443
ok: whisk auth set. Run 'wsk property get --auth' to see the new value.
ok: whisk API host set to https://sciabarra.net:443
ok: whisk namespace set to guest
```

sciabarracom:
ok the problem is not related to the deployment

PR #394, opened 2018-12-14
version bump all components to Dec 13 tags.

dgrove-oss:


dgrove-oss:
ready to merge -- I rebased to pick up php 7.3.

PR #397, opened 2018-12-16
Add PHP 7.3 runtime

akrabat:
Add support for the new PHP 7.3 runtime as per https://github.com/apache/incubator-openwhisk/issues/4174

### DIFF
--- a/helm/openwhisk/runtimes.json
+++ b/helm/openwhisk/runtimes.json
@@ -155,7 +155,7 @@
             },
             {
                 "kind": "php:7.2",
-                "default": true,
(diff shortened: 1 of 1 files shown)


helm/openwhisk/runtimes.json
```diff
@@ -166,6 +166,20 @@
                     "attachmentName": "codefile",
                     "attachmentType": "text/plain"
                 }
+            },
```
rabbah: change default above to false.
I wonder if this should be flagged as a bug in the controller when more than one kind is set as default.

PR #410, opened 2019-01-14
Update sample yaml snippet for external providers db

### LINKED ISSUE #407: error in documenting providers database
it is NOT

```
providers:
   alarm:
     enable: true
     db:
       ....
```

but

```
providers:
  db:
    ...
  alarm:
    enabled: true
```
For deploying with an external ingress you need

```
ingress:
   type: Standard
   domain: ingress.public.domain
   apiHostName: ingress.public.domain
```

it is not clearly documented (I had to read the code)

dgrove-oss:
Partially fixes #407

### DIFF
--- a/docs/configurationChoices.md
+++ b/docs/configurationChoices.md
@@ -42,18 +42,16 @@ user action containers created by the DockerContainerFactory are not configured 
 themselves be able to invoke Kubernetes services). To work around this you must do one
 of the following three alternatives:
 1. Deploy a CouchDB instance external to your Kubernetes cluster and configure the event
-provider(s) to use it by adding stanzas like the following to your `mycluster.yaml`:
+providers to use it by adding a stanza like the following to your `mycluster.yaml`:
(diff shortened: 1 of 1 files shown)


sciabarracom:
Yes but you have also to say you have to enable the providers. Also it is pretty important to document the "domain" value because otherwise the ingresses defaults to a value "domain" and nothing works...

dgrove-oss:
I think what is missing is the analogue to 
https://github.com/apache/incubator-openwhisk-deploy-kube/blob/master/docs/k8s-ibm-public.md
or 
https://github.com/apache/incubator-openwhisk-deploy-kube/blob/master/docs/k8s-google.md
for deploying on a roll-your-own cluster.  It's not specific to providers per se, but to specifying the Ingress that is appropriate to your Kubernetes solution.

dgrove-oss:
ready to merge -- fixes a doc bug @sciabarracom reported.

PR #411, opened 2019-01-14
add delay in healthy invoker readiness check

dgrove-oss:
In TravisCI environments, we often see failures in jobs that
run "right after" there is an initial report of healthy invokers
by the controller.  Let's wait for a few seconds in the readiness
probe to see if it improves the odds of success of jobs that run
triggered on this readiness probe.


dgrove-oss:
Did not help; abandon.

PR #414, opened 2019-01-15
adding troubleshooting for install packages failure

danielwhatmuff:


danielwhatmuff:
Hey @dgrove-oss thanks for merging. 

I did encounter a few issues with the install of openwhisk on kubernetes so have been taking a look through the installation process which runs upon a `helm install` and noticed there is quite a lot going on as part of `couchdb-init-job.yaml` and `install-packages-job.yaml` - for example the [myTask.sh](https://github.com/apache/incubator-openwhisk-deploy-kube/blob/master/helm/openwhisk/configMapFiles/installPackages/myTask.sh) script clones 5 repos from Github and executes a further 6 bash scripts. Also [initdb.sh](https://github.com/apache/incubator-openwhisk-deploy-kube/blob/master/helm/openwhisk/configMapFiles/initCouchDB/initdb.sh) clones another repo and runs 4 ansible playbooks.

I wondered if it would be possible to move any of this stuff into the build and release phase so that the helm install can just pull prebaked containers from a registry and either inject environment variables or mount configmaps to pods. With the aim being that provisioning openwhisk on kube is faster and more reliable.

Interested to hear your thoughts.

dgrove-oss:
Until relatively recently, the helm chart deployed the `latest` versions of all the components.  So, cloning the git repos in the install jobs made sense (it would have been messy to generate a new docker image every time one of the 5 upstream repos changed).   Now that we've switched to deploying pinned versions of all the upstream repos/images, it probably does make sense to publish an image that contains the pre-cloned source code from all the upstream repos and run the various jobs using that as a base image.  

I can take a look at doing that.

danielwhatmuff:
Makes sense. Thanks a lot.

dgrove-oss:
opened #431 so this idea doesn't get lost

PR #442, opened 2019-03-11
Get dind-cluster.sh from kubeadm-dind 0.1 release

dgrove-oss:


dgrove-oss:
@csantanapr or @rabbah -- could one of you approve and merge.  Hacks around an upstream problem with the dind-cluster script for kube 1.12.  Thanks.

PR #447, opened 2019-04-03
Add Support to store Openwhisk Component Images in Hosted Docker Registry.

neerajmangal:
## Description 

Currently, Private/Hosted Docker registry values are only used for installing the runtime images from playbook.yml on invoker pod. 

This PR is to support a scenario where we need to use our Hosted Docker registry to store Openwhisk Images as well for like invoker, controller, etc and only allowed to use Hosted Docker Repository (Artifactory) by Kubernetes Cluster provider (in our case Hosted Kubernetes Deployment). 

Furthermore, it will provide a more configurable option to use image location in pod configs.
 
## Environment details:

* On Premise Hosted Private Kuberentes Cluster
* Images are uploaded to Private/Hosted Docker Registry   

## Steps to reproduce the issue:

1. Hosted Private Kubernetes provider only allows using Hosted Docker Registry (Like Artifactory).     
2. Current Helm deployment does not allow to use Private Docker registry for Openwhisk Component images. as of today, manual changes required in pod/job yaml files to use private Docker registry host/server. 

# Solution 

## Prerequisite - 
- Upload all images of Openwhisk in private/hosted docker repository server. 

## Changes 

- Define an imagePullSecret
- Use a Value Flag to use imagePullSecrets and image repo config. 
- Create a secret of type kubernetes.io/dockerconfigjson for private/hosted docker registry

### DIFF
--- a/helm/openwhisk/templates/_invoker-helpers.tpl
+++ b/helm/openwhisk/templates/_invoker-helpers.tpl
@@ -33,11 +33,7 @@
 
 {{- define "openwhisk.docker_pull_runtimes" -}}
 - name: docker-pull-runtimes
-{{- if .Values.docker.usePrivateRegistry }}
   image: "{{- .Values.docker.registry.name -}}{{- .Values.utility.imageName -}}:{{- .Values.utility.imageTag -}}"
-{{- else }}
-  image: "{{- .Values.utility.imageName -}}:{{- .Values.utility.imageTag -}}"
-{{- end }}
   imagePullPolicy: {{ .Values.utility.imagePullPolicy | quote }}
   command: ["/usr/local/bin/ansible-playbook", "/invoker-scripts/playbook.yml"]
   volumeMounts:
--- a/helm/openwhisk/templates/_invoker-helpers.tpl
+++ b/helm/openwhisk/templates/_invoker-helpers.tpl
@@ -33,8 +33,12 @@
 
 {{- define "openwhisk.docker_pull_runtimes" -}}
 - name: docker-pull-runtimes
-  imagePullPolicy: {{ .Values.utility.imagePullPolicy | quote }}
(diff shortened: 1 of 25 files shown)


helm/openwhisk/templates/_invoker-helpers.tpl
```diff
@@ -33,8 +33,12 @@
 
 {{- define "openwhisk.docker_pull_runtimes" -}}
 - name: docker-pull-runtimes
-  imagePullPolicy: {{ .Values.utility.imagePullPolicy | quote }}
+{{- if .Values.docker.usePrivateRegistry }}
```
dgrove-oss: Why put this in an if/else block instead of letting .Values.docker.registry.name default to an empty string?

neerajmangal: Hi @dgrove-oss, I used this based on the usePrivateRegistry flag keeping a scenario where the user won't like to use private Docker registry for component images but want to use only for runtime images (like today's behavior).

Though using your method is better, in that case, one has to put all component + runtime images in Private Docker Registry is registry name is provided. Let me know your thoughts, I will modify it based on your suggestion.

neerajmangal: Hi @dgrove-oss, I modified it based on your suggestion and removed an extra flag(usePrivateRegistry). Now if docker registry is not empty, then charts will use imagePullSecrets and docker registry provided to pull the images. Let me know if you have further suggestions.

dgrove-oss: This looks reasonable to me.  Is it still a WIP or is it ready for merging?

neerajmangal: Thanks @dgrove-oss, it is ready to be merged now.

Issue #469, opened 2019-05-29
upstream zookeeper image change broke liveness/readiness probe


dgrove-oss:
zookeeper's official dockerhub image changed its base image from `openjdk:8-jre-alpine` to `openjdk:8-jre-slim` 4 days ago in commit https://github.com/31z4/zookeeper-docker/commit/01d134a4dcb1676d7efe9d0c64c5e0cfdba6fb27

This removed the `nc` command from the image, which breaks the liveness and readiness probes for the zookeeper pod.

dgrove-oss:
I opened https://github.com/31z4/zookeeper-docker/issues/66 to inquire if they would add `nc` back to the base image.

dgrove-oss:
nc was added to the upstream image in https://github.com/docker-library/official-images/pull/6005

PR #471, opened 2019-05-30
disable zookeeper readiness probe

dgrove-oss:
Comment out zookeeper readiness probe until we have guidance
from upstream project about whether they will restore `nc`
to their image or if we will need to develop an alternative
readiness probe.

See #469.

### DIFF
--- a/helm/openwhisk/templates/zookeeper-pod.yaml
+++ b/helm/openwhisk/templates/zookeeper-pod.yaml
@@ -62,14 +62,15 @@ spec:
             port: {{ .Values.zookeeper.port }}
           initialDelaySeconds: 5
           periodSeconds: 10
-        readinessProbe:
-          exec:
-            command:
-            - /bin/bash
-            - -c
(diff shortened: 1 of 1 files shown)


neerajmangal:
Hi @dgrove-oss, I was thinking on instead of disabling of probes we can have configurable probes as I faced similar issues with other components and has to manually configure them. I am working on another PR to do that for other components as well like controller etc. WDYT?

dgrove-oss:
Hi @neerajmangal, would be interested in seeing the PR and discussing.  Ideally the probes are a low-level detail that we can get to always work without bothering the user.  If we can't, figuring out an understandable way to configure problematic ones them does seem useful.  I'd hope the low-level probes that are just hitting an http or tcp endpoint at a specific container port would not need configuration.

neerajmangal:
Agree with you @dgrove-oss. My bad, I was trying to refer to probes timing setting only like `initialDelaySeconds` etc to be configurable, not the actual low-level `tcpSocket` or `httpGet` probes.

PR #482, opened 2019-06-18
Add configurations for prometheus and metrics

daisy-ycguo:
This PR is to add configurations for prometheus and metrics supporting.

After adding below configurations, controller and invoker are exposing `/metrics` endpoints on container port `8080`.
```
metrics:
  prometheusEnabled: true
  whiskconfigFile: "whiskconfig.conf"
  kamonEnabled: true
  kamonTags: true
```

### DIFF
--- a/helm/openwhisk/templates/invoker-pod.yaml
+++ b/helm/openwhisk/templates/invoker-pod.yaml
@@ -169,7 +169,22 @@ spec:
           # properties for zookeeper connection
           - name: "ZOOKEEPER_HOSTS"
             value: "{{ include "openwhisk.zookeeper_connect" . }}"
-
+{{ if .Values.metrics.prometheusEnabled }}
+          - name: "OPENWHISK_ENCODED_CONFIG"
+            value: {{ template "openwhisk.whiskconfig" . }}
+{{ end }}
+{{ if or .Values.metrics.kamonEnabled .Values.metrics.prometheusEnabled }}
+          - name: "METRICS_KAMON"
+            value: "true"
+{{ end }}
+{{ if and .Values.metrics.kamonTags .Values.metrics.prometheusEnabled }}
+          - name: "METRICS_KAMON_TAGS"
+            value: "true"
+{{ end }}
+{{ if .Values.metrics.userMetricsEnabled }}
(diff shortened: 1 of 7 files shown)


helm/openwhisk/templates/invoker-pod.yaml
```diff
@@ -169,7 +169,22 @@ spec:
           # properties for zookeeper connection
           - name: "ZOOKEEPER_HOSTS"
             value: "{{ include "openwhisk.zookeeper_connect" . }}"
-
+{{ if .Values.metrics.prometheusEnabled }}
+          - name: "OPENWHISK_ENCODED_CONFIG"
```
neerajmangal: Just a suggestion, instead of having OPENWHISK_ENCODED_CONFIG can we have Config_whisk_metrics_prometheusEnabled env to true, if that works. As OPENWHISK_ENCODED_CONFIG env is general purpose typesafe config file but here it seems only applicable when `.Values.metrics.prometheusEnabled `. 

In the future, we may think of having all env configs generated from values via templates and then passed for each component separately using OPENWHISK_ENCODED_CONFIG for consistency.

daisy-ycguo: I agree with you. But currently the implementation of prometheus configuration in OpenWhisk core project is using OPENWHISK_ENCODED_CONFIG. See https://github.com/apache/incubator-openwhisk/pull/4309 . So I have to use OPENWHISK_ENCODED_CONFIG here.

Maybe you could raise an issue to OpenWhisk core project. After they change, we can change it here.

daisy-ycguo:
@dgrove-oss updated. Please review again. Thanks.

helm/openwhisk/templates/invoker-pod.yaml
chetanmeh: @daisy-ycguo encoded config is more for convenience. So you can avoid that for now (and we later do a wholesale move to that for all other configs) and use what @neerajmangal suggested i.e. `Config_whisk_metrics_prometheusEnabled` which should have same effect.

Above PR was more to simplify using the encoded config approach with ansible based setups

PR #491, opened 2019-07-03
Use most recent release for runtimes and apigateway

dgrove-oss:


helm/openwhisk/runtimes.json
```diff
@@ -35,7 +35,7 @@
                 "image": {
                     "prefix": "openwhisk",
                     "name": "action-nodejs-v10",
-                    "tag": "ab224ab"
+                    "tag": "1.14.0-incubating"
```
rabbah: 1.14 not 1.13?

dgrove-oss: we did a 1.14.0 release of nodejs a couple weeks ago.

PR #494, opened 2019-07-05
Support Pod Disruption Budget

### LINKED ISSUE #492: Enhancement : Adding Pod Disruption Budget for Openwhisk Deployment
Placeholder issue for implementing and discussing [Pod Disruption budgets](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/). 

Wanted to start implementing PDB for zookeeper, invoker, and controller pods so that they can survive from [voluntary and involuntary disruptions](https://kubernetes.io/docs/tasks/run-application/configure-pdb/). 

```yaml
values.yaml 

pdb:
  enabled: true # can be put in all component separately if we want component specific PDB
  zookeeper: 
      maxUnavailable:  1 # It can be % or can be based on minAvailable as well. 
  controller:
      maxUnavailable: 1
  invoker:
      maxUnavailable: 1
```

```yaml
zookeeper-pdb.yaml

{{ if gt (int .Values.zookeeper.replicaCount) 1 }}
....
....
spec:
  selector:
    matchLabels:
      name: "{{ .Release.Name }}-zookeeper"
  maxUnavailable: {{.Values.pdb.zookeeper.maxUnavailable}}
```

Please provide your feedback on above, will raise a PR soon.

neerajmangal:
Fixes #492 

Implemented PDB for below components
- Zookeeper
- Kafka
- Controller
- Invoker

### DIFF
--- a/helm/openwhisk/values-metadata.yaml
+++ b/helm/openwhisk/values-metadata.yaml
@@ -1595,7 +1595,7 @@ pdb:
       description: "Number of pods from replica set that can be unavailable after the eviction. Should be an absolute number and less than replicaCount value"
       type: "number"
       required: false
-  Kafka:
+  kafka:
     _metadata:
     label: "Kafka"
     maxUnavailable:
--- a/helm/openwhisk/values-metadata.yaml
+++ b/helm/openwhisk/values-metadata.yaml
@@ -1500,12 +1500,28 @@ affinity:
 
 probes:
   _metadata:
-  label: "Pod's liveness and readiness time related settings"
+  label: "Pod's Probes Settings"
   description: "Used for configurable liveness and readiness probes timing settings"
-  component:
-    _metadata:
-      label: "Specify component name i.e. zookeeper, controller etc."
-      description: "Specify component name for which probes settings will apply"
+  controller:
+    initialDelaySeconds:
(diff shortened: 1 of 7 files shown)


helm/openwhisk/values-metadata.yaml
```diff
@@ -1560,4 +1560,24 @@ metrics:
...
+  _metadata:
+  label: "Pod's disruption budget settings"
+  description: "Pod Disruption settings for voluntary and nonvoluntary disruptions"
+  enable:
+    _metadata:
+      label: "Enable pod disruption budget for deployment"
+      description: "true to enable pod disruption budgets for pods"
+      type: "boolean"
+      required: true
+  component:
```
dgrove-oss: The tool that processes values-metadata.yaml won't understand the pattern you are describing here by using `component`.   I think the choices are either (a) replicate the component name & maxUnavailable stanza for each component (zookeeper, controller, etc.) that you have defined in values.yaml or (b) leave off the entries in values-metadata.yaml.    I'm fine with either option.

neerajmangal: Sure will do that. Also, will do a similar modification for prob settings as well.

PR #506, opened 2019-08-07
use extraEnvVars to set __OW_ALLOW_CONCURRENT

dgrove-oss:


dgrove-oss:
Picks up the fix for https://github.com/apache/openwhisk/issues/4569 and changes helm chart accordingly.

Issue #512, opened 2019-08-29
helm install .... --name owdev-name-test   can not have "-" in it


ChangZihao:
I deploy the openwhisk with helm in minikube on my ubuntu 18.04 server.
If I set the --name  with '-'   like owdev-name-test, the kafka pod will start with error

```
waiting for kafka to be ready
[2019-08-29 02:39:23,810] FATAL  (kafka.Kafka$)
org.apache.kafka.common.config.ConfigException: Invalid value test for configuration broker.id: Not a number of type INT
	at org.apache.kafka.common.config.ConfigDef.parseType(ConfigDef.java:713)
	at org.apache.kafka.common.config.ConfigDef.parseValue(ConfigDef.java:460)
	at org.apache.kafka.common.config.ConfigDef.parse(ConfigDef.java:453)
	at org.apache.kafka.common.config.AbstractConfig.<init>(AbstractConfig.java:62)
	at kafka.server.KafkaConfig.<init>(KafkaConfig.scala:883)
	at kafka.server.KafkaConfig$.fromProps(KafkaConfig.scala:867)
	at kafka.server.KafkaConfig$.fromProps(KafkaConfig.scala:864)
	at kafka.server.KafkaServerStartable$.fromProps(KafkaServerStartable.scala:28)
	at kafka.Kafka$.main(Kafka.scala:58)
	at kafka.Kafka.main(Kafka.scala)
```

My helm command is 
``` bash 
helm install ~/projects/openwhisk-deploy-kube/helm/openwhisk --namespace=openwhisk-kube --name=owdev-name-test -f mycluster-name-test.yaml
```
mycluster-name-test.yaml
```
whisk:
    ingress:
        type: NodePort
        apiHostName: 192.168.99.110
        apiHostPort: 31001

nginx:
    httpsNodePort: 31001

invoker:
    containerFactory:
        impl: "kubernetes"
```
Now I have found the bug , and fixed it

dgrove-oss:
Fixed by #513

PR #513, opened 2019-08-29
fix: kafka, zookeeper, controller get num in hostname

ChangZihao:
#512


chetanmeh:
Thanks for the fix. This thing hit us also once when we used `-` in release names. Good to have a proper fix for that

Issue #436, opened 2019-02-25
Api Gateway not working on Docker Desktop Windows/Mac


andreialecu:
Api Gateway seems to be broken with the kube chart. There must be some internal networking problem. 

```
$ wsk -i action invoke hello --result
{
    "payload": "Hello, World!"
}

$ wsk -i api create /hello get hello --response-type json
ok: created API /hello GET for action /_/hello
https://192.168.65.3:31001/api/23bc46b1-71f6-4ed5-8c54-816aa4f8c502/hello

$ curl -k https://localhost:31001/api/23bc46b1-71f6-4ed5-8c54-816aa4f8c502/hello
{"code":502, "message":"Oops. Something went wrong. Check your URI and try again."}
```

```
==> /var/log/api-gateway/gateway_error.log <==
2019/02/25 15:05:18 [error] 38#0: *28 connect() failed (111: Connection refused) while connecting to upstream, client: 10.1.0.96, server: , request: "GET /api/23bc46b1-71f6-4ed5-8c54-816aa4f8c502/hello HTTP/1.1", upstream: "https://127.0.0.1:31001/api/v1/web/guest/default/hello.json", host: "owdev-apigateway.openwhisk.svc.cluster.local:8080"
2019/02/25 15:05:18 [warn] 38#0: *28 upstream server temporarily disabled while connecting to upstream, client: 10.1.0.96, server: , request: "GET /api/23bc46b1-71f6-4ed5-8c54-816aa4f8c502/hello HTTP/1.1", upstream: "https://127.0.0.1:31001/api/v1/web/guest/default/hello.json", host: "owdev-apigateway.openwhisk.svc.cluster.local:8080"
2019/02/25 15:05:18 [crit] 38#0: *28 connect() to [::1]:31001 failed (99: Address not available) while connecting to upstream, client: 10.1.0.96, server: , request: "GET /api/23bc46b1-71f6-4ed5-8c54-816aa4f8c502/hello HTTP/1.1", upstream: "https://[::1]:31001/api/v1/web/guest/default/hello.json", host: "owdev-apigateway.openwhisk.svc.cluster.local:8080"
[...]

PR #518, opened 2019-08-30
update to use apigateway-0.11.0

dgrove-oss:


dgrove-oss:
Pre-req to picking up localmode fix to enable addressing #436

PR #536, opened 2019-10-07
kind: describe using `kind load` to deploy locally-built image

dgrove-oss:


docs/k8s-kind.md
```diff
@@ -112,6 +112,31 @@ we will use the port forwarding configured by the `extraPortMappings`
...
+```
+Then add a stanza to your `mycluster.yaml` to override the default
+behavior of pulling a stable `openwhisk/controller` image from Docker Hub.
+```yaml
+controller:
+  imageName: "whisk/controller"
+  imageTag: "latest"
+```
+
+Then deploy OpenWhisk normally using `helm install`. The deployed
```
chetanmeh: Should we document using `helm upgrade --install` instead of install for incremental deployment to avoid creating new release and just refresh the existing controller/invoker?

dgrove-oss: we should probably add some documentation on options for selective restarting of individual containers.  going to push that off to a future PR.

PR #568, opened 2020-01-06
NFS Dynamic Provisioning setup guide 

giusdp:
I went more in depth on how to setup nfs dynamic provisioning, by explaining what to do to setup the nfs server and the nfs client.

I don't know if it's overkill, especially the manual nfs client, but I often had problems using the helm charts and had to revert to do it manually.

Here I wrote down guides and videos that I myself followed for my openwhisk deployment. I thought having it all in one place could be useful for others. Let me know what you think about it!

### DIFF
--- a/docs/k8s-nfs-dynamic-storage.md
+++ b/docs/k8s-nfs-dynamic-storage.md
@@ -21,14 +21,53 @@
 
 ## NFS-based Dynamic Provisioning
 
-You will need an already-provisioned NFS server supporting NFS v4 or better,
-preferably provisioned for at least 5 GB.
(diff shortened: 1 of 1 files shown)


dgrove-oss:
Could you file an Apache CLA as described in https://github.com/apache/openwhisk/blob/master/CONTRIBUTING.md?  This is a big enough contribution that we should get the legalities taken care of.

giusdp:
Just did it. Hope it gets accepted soon

giusdp:
The cla has been filed, the id I used is "giusdp". Now all should be good to keep contributing :smile:

PR #584, opened 2020-02-14
NetworkPolicy to isolate pods running user actions

dgrove-oss:
Add NetworkPolicy to enforce moderate levels of network isolation:
  1. User action pods only accept incoming connections from invokers
  2. Back-end control plane pods only accept incoming connections
     from other control plane pods.


neerajmangal:
Thanks, @dgrove-oss. LGTM.

PR #600, opened 2020-05-07
docs fix: Helm v3 doesn't automatically create the target namespace

dgrove-oss:


README.md
```diff
@@ -230,7 +230,12 @@ For simplicity, in this README, we have used `owdev` as the release name and
 You can use a different name and/or namespace simply by changing the commands
 used below.
 
-Deployment can be done by using the following single command:
+First, create the target namespace for your OpenWhisk deployment:
+```shell
+kubectl create namespace openwhisk
+```
+
+Next, deploy openwhisk to that namespace:
```
rabbah, line 238:
```suggestion
Next, deploy OpenWhisk to that namespace:
```

wkk:
There is a new option in helm v3.2.0: `--create-namespace` which allow helm 3 to create the namespace if it doesn't exist (like the default behavior in helm 2)

The PR that added this option:
https://github.com/helm/helm/pull/7648

Using this command we can still deploy using a single command:

```shell
helm install owdev ./helm/openwhisk -f mycluster.yaml -n openwhisk --create-namespace
```

dgrove-oss:
I changed to use @wkk suggestion of adding `--create-namespace` to a single command line.

PR #613, opened 2020-06-19
Fix issue with quoted docker registry credentials

mluds:
The docker registry credentials are quoted here:
https://github.com/apache/openwhisk-deploy-kube/blob/bbf7846517ed690bb0bf2fdda444222a82037aa8/helm/openwhisk/templates/ow-docker-registry-secret.yaml#L26-L27

When used in the dockerPullRuntimes Ansible, literal quotes are passed in the username and password, breaking authentication.

This fixes that by stripping these before use.

### DIFF
--- a/helm/openwhisk/configMapFiles/dockerPullRuntimes/playbook.yml
+++ b/helm/openwhisk/configMapFiles/dockerPullRuntimes/playbook.yml
@@ -25,8 +25,8 @@
     docker_pull_retries: 10
     docker_pull_delay: 10
     runtimes_registry: "{{ lookup('env', 'RUNTIMES_REGISTRY') | default() }}"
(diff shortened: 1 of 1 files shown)


dgrove-oss:
I think quoting the values before running them through b64enc is a mistake.  We don't do that anywhere else.   I'll submit a PR to take out the `| quote` as an alterantive

dgrove-oss:
I think #616 is a better fix..let's keep the bogus quotes out of the secret

dgrove-oss:
@mluds - I think #616 should fix this.  I merged it this morning.  When you get a  chance, could you give it a try and let us know if I missed something?  Thanks

mluds:
Sorry for the late response, but I just tried this and looks like it's working. Thanks!

PR #615, opened 2020-06-26
Add flexibility to configure grafana

possani:
# Context

I am experimenting with Openwhisk and Openstack and currently using _kubectl proxy_ to access the dashboard. However, for that to work, I have two options:

1. Change de variable on the fly:
`kubectl -n openwhisk set env deployment/owdev-grafana GF_SERVER_ROOT_URL=http://localhost:8001/api/v1/namespaces/openwhisk/services/http:owdev-grafana:http/proxy/`

2. Replace the value in the manifest before installing the chart.

# What would you expect?

To be able not only to set GF_SERVER_ROOT_URL but any other environment variable in the configuration file.

`helm install owdev openwhisk/helm/openwhisk -n openwhisk --create-namespace -f mycluster.yaml`

### DIFF
--- a/helm/openwhisk/values.yaml
+++ b/helm/openwhisk/values.yaml
@@ -335,6 +335,18 @@ grafana:
   - https://raw.githubusercontent.com/apache/openwhisk/master/core/monitoring/user-events/compose/grafana/dashboards/openwhisk_events.json
   - https://raw.githubusercontent.com/apache/openwhisk/master/core/monitoring/user-events/compose/grafana/dashboards/global-metrics.json
   - https://raw.githubusercontent.com/apache/openwhisk/master/core/monitoring/user-events/compose/grafana/dashboards/top-namespaces.json
+  # https://grafana.com/docs/grafana/latest/administration/configuration/#configure-with-environment-variables
+  env:
+  - name: "GF_AUTH_ANONYMOUS_ENABLED"
+    value: "true"
+  - name: "GF_PATHS_PROVISIONING"
+    value: "/var/lib/grafana/config"
+  - name: "GF_USERS_ALLOW_SIGN_UP"
+    value: "false"
+  - name: "GF_SERVER_ROOT_URL"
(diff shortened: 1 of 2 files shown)


possani:
Hi @dgrove-oss, who could take a look at my PR? The idea would be to move the environment variables from the deployment template to the values file to give it more flexibility.

helm/openwhisk/values.yaml
```diff
@@ -335,6 +335,10 @@ grafana:
   - https://raw.githubusercontent.com/apache/openwhisk/master/core/monitoring/user-events/compose/grafana/dashboards/openwhisk_events.json
   - https://raw.githubusercontent.com/apache/openwhisk/master/core/monitoring/user-events/compose/grafana/dashboards/global-metrics.json
   - https://raw.githubusercontent.com/apache/openwhisk/master/core/monitoring/user-events/compose/grafana/dashboards/top-namespaces.json
+  # https://grafana.com/docs/grafana/latest/administration/configuration/#configure-with-environment-variables
+  env:
+  - name: "TEST"
```
neerajmangal, line 340: I think now, it would be good to move the default env to this section in values.yaml.

possani: I actually did that when I first created the MR. The only variable I had left was GF_SECURITY_ADMIN_PASSWORD because it already uses another field in the values.yaml file, but one of the tests failed. I'll push that commit again.

possani:
```system.basic.WskActionTests > Whisk actions should pass parameters bound on creation-time to the action STANDARD_ERROR
    org.scalatest.exceptions.TestFailedException: "2020-06-26T19:27:03.247708024Z stdout: params.param1: test1 2020-06-26T19:27:03.247795526Z stdout:  params.param2: test2" did not include substring "params.payload: this is a test"

possani:
This should've been closed a while ago.

PR #625, opened 2020-07-27
docker login on invoker pods when using private registry

mluds:
Actions that used a custom image from a private registry were failing, since the invoker containers weren't logged into the registry. This commit runs `docker login` in the invoker containers on startup to fix that.


dgrove-oss:
This looks reasonable to me.  @mluds, please comment if you are done (not still working on this) and are ready for a potential merge.  Do you have an Apache CLA on file?

PR #639, opened 2020-09-18
Add helm release to invoker pod

ningyougang:
If want to deploy two helm release openwhisk cluster,  when the second helm release cluster's invokerN pod is restarting, it will delete the first helm release's invokerN runtime pods.

Because when invokerN pod is starting, it will delete  runtime pods with label(invoker=invokerN) only. e.g.
https://github.com/apache/openwhisk/blob/master/core/invoker/src/main/scala/org/apache/openwhisk/core/containerpool/kubernetes/KubernetesContainerFactory.scala#L65

in order to avoid `delete other helm release invokerN's runtime pod`, it is better to add helm release to invoker pod, and when invoker pod is starting, add label(release=xxxxxx) together with label(invoker=invokerN) to delete runtime pods.

another brother pr: https://github.com/apache/openwhisk/pull/4979

### DIFF
--- a/helm/openwhisk/templates/invoker-pod.yaml
+++ b/helm/openwhisk/templates/invoker-pod.yaml
@@ -205,6 +205,8 @@ spec:
           - name: "CONFIG_whisk_spi_ActivationStoreProvider"
             value: "org.apache.openwhisk.core.database.elasticsearch.ElasticSearchActivationStoreProvider"
 {{- end }}
+          - name: "CONFIG_whisk_helm_release"
+            value: "{{ .Release.Name }}"
         ports:
         - name: invoker
           containerPort: {{ .Values.invoker.port }}


dgrove-oss:
Repeating the question from the core repo....does deploying to separate namespaces avoid the problem?  My intuition is that it should.

ningyougang:
@dgrove-oss ,hi
yes, in two kubernateds namespace can avoid the problem.
but some users may deploy two deployments in same namespace.

hm..i think add a one more lable(release=xxxx) it is more carefully.

dgrove-oss:
I don't object to much to the change, but be aware that when you do a `helm delete` and undeploy the chart, it doesn't remove the action containers (the chart had no idea they exist).  So using a namespace for more than a single deployment is a bit dicey in any case.  Part of the reason the invoker deletes all the pods when it starts up is to try to clean up from previous deployments in the same namespace.

dgrove-oss (approved):

PR #650, opened 2020-10-26
Updated to include Windows tools

edokan:
Instead of `grep` which is not deployed with Windows by default, converted code to use built-in `find` tool.
Also added a step to reduce possible installation issues

### DIFF
--- a/docs/k8s-docker-for-windows.md
+++ b/docs/k8s-docker-for-windows.md
@@ -60,7 +60,7 @@ the box to enable Kubernetes.
 ### Configuring OpenWhisk
 
 You will be using a NodePort ingress to access OpenWhisk. Assuming
-`kubectl describe nodes | grep InternalIP` returns 192.168.65.3 and
+`kubectl describe nodes | find "InternalIP"` returns 192.168.65.3 and
 port 31001 is available to be used on your host machine, a
 mycluster.yaml for a standard deployment of OpenWhisk would be:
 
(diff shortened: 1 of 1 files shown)


rabbah (approved):
Thanks!!

docs/k8s-docker-for-windows.md
```diff
@@ -77,6 +77,10 @@ nginx:
 
 ### Using helm to install OpenWhisk
 
+Installation expects `openwhisk` namespace to be created. To create, run
```
dgrove-oss, line 80: What version of Helm3 does one get on Windows?  If it supports `--create-namespace` then this can be rolled into the helm install command:
```
helm install owdev ./helm/openwhisk -n openwhisk -create-namespace -f mycluster.yaml
```

edokan: I've just installed and its version is 3.1.2 . `helm help install` does not list `--create-namespace` among the flags. Being a newcomer on everything docker related, I've probably would never found that either. So I've included what worked, and probably will work for everyone to save a few like me from a google search.

dgrove-oss: ok.  My recollection is that --create-namespace was added in helm 3.2.  So if that's not the default in windows yet it is better to include the extra step here.

dgrove-oss (approved):
Thanks for the contribution!

docs/k8s-docker-for-windows.md
edokan: it is my bad. I had an old installation of helm in my path, so I used old version 3.1.2, instead of newer version that comes with kubernetes-helm (3.3.4). Do you want me to update documentation?

dgrove-oss: I merged already, so would need a new PR to take out the paragraph about creating the namespace.  Would be great if you had the time to do that, but not critical.

PR #651, opened 2020-11-12
Make busybox image configurable

jiangpengcheng:
when using a private docker registry, users may not have access to the root path, so they can not push busybox image to `xxx.xxx.xxx/busybox`

this PR makes busybox's image configurable to resolve above issue


dgrove-oss (approved):
LGTM.  Thanks!

PR #653, opened 2020-11-18
Add timeout configuration

ningyougang:
Fix https://github.com/apache/openwhisk-deploy-kube/issues/652

In kubenetes env,  if runtime pod number is big and `akka.coordinated-shutdown.phases.actor-system-terminate_timeout` is low(e.g default value is `30 s`), this may lead to `the runtime pods are not deleted completely after execute helm delete/uninstall xxx`, because when timeout happens, the `cleanup`  is not executed finished.

In such a case, need to increase relative timeout value, it is better to make these timeout value configurable.
- [x] actorSystem terminate timeout configuration
- [x] runtime pods delete timeout configuration

Another brother pr: https://github.com/apache/openwhisk/pull/5028

### DIFF
--- a/helm/openwhisk/templates/invoker-pod.yaml
+++ b/helm/openwhisk/templates/invoker-pod.yaml
@@ -207,6 +207,10 @@ spec:
 {{- end }}
           - name: "CONFIG_whisk_helm_release"
             value: "{{ .Release.Name }}"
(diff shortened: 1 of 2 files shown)


dgrove-oss (approved):
LGTM

jiangpengcheng (approved):

PR #658, opened 2020-12-14
Fix invalid kamonTags setting

upgle:
## Description
There was an issue that the invoker's kamonTags setting was not applied. 
On the other hand, the controller had no problem.

I've found a problem with the conditional and fixed it.
(The kamon tag setting is independent of Prometheus.)


style95 (approved):

dgrove-oss:
ok to merge this?

style95:
I merged it.

PR #664, opened 2021-01-08
add-scc-to-user to owdev-init-sa

unreal-altran:
To Avoid Write permission on cert folder....

Tested on OKD/OpenShift 4.5 works well...


dgrove-oss (approved):
Thanks for the contribution!

unreal-altran:
I have also replay to this bug on the exactly same problem.
https://github.com/apache/openwhisk-deploy-kube/issues/594

PR #668, opened 2021-01-25
Optimize nginx configuration

ningyougang:
- Add nginx worker_processes
After changed to `auto`, the performance increases a lot during benchmark
- Add upstream section
Before changed to upsteam, the tps under our env is
image
After added the upsteam section, the tps is
image
controller replicaCount: `2`, invoker replicaCount: `4`, benchmark 10 hello actions at the same time.
(BTW, when i benchmarked this point, i didn't add `worker_processes: auto`)

Above 2 configuration can impove the performance both


style95:
For the record, the benchmark results came from our downstream version.
So with the upstream codebase, there might be some differences.
But still, we believe the upstream codes can benefit from this change.

dgrove-oss (approved):

PR #716, opened 2021-11-11
Update helm to make secret instead of configmap for Nginx External Cert usage

paul42:
In my quest to use my own certs with an internal CA, and with some great help from @style95 I was able to make significant headway and get my deployment working as desired. I did want to make a small tweak so the external cert for nginx is not stored as a configmap, but instead as a secret. (I realize there's not a ton of difference under the hood but it helps me sleep better at night) I have not tested this with any other deployment types, so feel free to give any feedback as needed.


dgrove-oss (approved):
This looks reasonable to me.

dgrove-oss:
Closing & reopening to kick travis

dgrove-oss:
Trying to kick travis again...

dgrove-oss:
Kicking travis

PR #725, opened 2022-03-01
affinity stanzas in provider pods

dgrove-oss:


hunhoffe:
Your changes work well to add affinity to the provider pods but I believe there's a bug in the _affinity.tpl file that prevents the provider pods from being scheduled on invoker nodes.

If you remove the line [here](https://github.com/dgrove-oss/openwhisk-deploy-kube/blob/250fa8b22259fb63cc040230eb321a1b02101709/helm/openwhisk/templates/_affinity.tpl#L76) then it appears to work (in my local setup, at least!)

style95 (approved):

PR #729, opened 2022-05-06
Deploy with new scheduler

hunhoffe:
What this PR contains:
- Deploy etcd and scheduler components
- Allow option to enable or disable the scheduler (similar to ansible)
- Akka cluster config setup for replicating the scheduler 
- Tested with and without scheduler enabled on 3-node Kubernetes cluster with CouchDB activation store

Some changes are needed to the scheduler to support the akka cluster discovery mechanism. The scheduler also needs to be integrated into the build process. These changes are in [PR#5278](https://github.com/apache/openwhisk/pull/5278) of apache/openwhisk.

Tasks:
- [x] Merge/finalize PR for ```apache/openwhisk``` with Akka cluster management changes for scheduler, controller & updates to scripts that create docker images, etc.
- [x] Update docker image tags once images are generated
- [x] Update documentation

Future Work:
- etcd will currently always start fresh as a deployment specifically for an openwhisk deployment; future work could be to add persistence and also to support external etcd. I added in documentation that persistence to etcd was not currently supported and in a comment that the external etcd was not supported.
-  Does not support dedicated invoker feature, e.g., https://github.com/apache/openwhisk/pull/5292


helm/openwhisk/values.yaml
```diff
@@ -248,8 +254,8 @@ nginx:
 
 # Controller configurations
 controller:
-  imageName: "openwhisk/controller"
-  imageTag: "c5970a6"
+  imageName: "hunhoffe/controller"
```
hunhoffe, line 257: TODO: need to update this once apache/openwhisk has accepted changes needed for controller to replicate properly

helm/openwhisk/values.yaml
```diff
@@ -259,6 +265,54 @@ controller:
   jvmOptions: ""
   loglevel: "INFO"
 
+# Scheduler configurations
+scheduler:
+  enable: true
+  imageName: "hunhoffe/scheduler"
```
hunhoffe, line 271: TODO: need to update this once apache/openwhisk has accepted changes needed for scheduler to replicate properly, and also once scheduler images are pushed to dockerhub.

hunhoffe:
Pushed but having issues with inProgressJobRetention parsing in new scheduler image, see: https://github.com/apache/openwhisk/pull/5281

helm/openwhisk/values.yaml
```diff
@@ -249,20 +255,69 @@ nginx:
...
+  restartPolicy: "Always"
+  port: 8080
+  options: ""
+  jvmHeapMB: "1024"
+  jvmOptions: ""
+  loglevel: "INFO"
+
+# Scheduler configurations
+scheduler:
+  enable: true
```
hunhoffe, line 270: Do we want this to default to true or not? I set it to true here so that the travis tests would run by default with the scheduler, but perhaps the correct thing to do is run travis tests both with and without the scheduler?

tools/travis/deploy-chart.sh
```diff
@@ -216,12 +216,17 @@ if [ "${OW_LEAN_MODE:-false}" == "false" ]; then
   # Wait for the controller to confirm that it has at least one healthy invoker
   verifyHealthyInvoker
 
+  # Wait for scheduler to be up
+  statefulsetHealthCheck "ow4travis-scheduler"
```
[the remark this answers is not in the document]
style95: If we would not actually invoke a health action, I think it would be enough to go.

helm/openwhisk/values.yaml
style95: +1 to David's opinion.

We may take the same approach with the core repo.
image

It has a dedicated CI pipeline for the new scheduler.

helm/openwhisk/templates/scheduler-pod.yaml
```diff
@@ -0,0 +1,273 @@
...
+      {{- end }}
+
+      {{- if .Values.toleration.enabled }}
+      tolerations:
+{{ include "openwhisk.toleration.core" . | indent 8 }}
+      {{- end }}
+
+      initContainers:
+      # The scheduler must wait for kafka and/or couchdb to be ready before it starts
+{{ include "openwhisk.readiness.waitForController" . | indent 6 }}
```
style95, line 62: Should this be `waitForKafka` and `waitForCouchDB`?

helm/openwhisk/templates/scheduler-pod.yaml
```diff
@@ -0,0 +1,273 @@
...
+          timeoutSeconds: {{ .Values.probes.scheduler.livenessProbe.timeoutSeconds }}
+        readinessProbe:
+          httpGet:
+            path: "/ping"
+            port: {{ .Values.scheduler.port }}
+            scheme: "HTTP"
+          initialDelaySeconds: {{ .Values.probes.scheduler.readinessProbe.initialDelaySeconds }}
+          periodSeconds: {{ .Values.probes.scheduler.readinessProbe.periodSeconds }}
+          timeoutSeconds: {{ .Values.probes.scheduler.readinessProbe.timeoutSeconds }}
+        env:
```
style95, line 95: I think the `whick_cluster_name` needs to be added to differentiate different versions of clusters.
https://github.com/apache/openwhisk/blob/master/ansible/roles/schedulers/tasks/deploy.yml#L223

style95: Similarly, would be great to add `kafka_topic_prefix` too.
https://github.com/apache/openwhisk/blob/master/ansible/roles/schedulers/tasks/deploy.yml#L150

BTW, it seems the prefix configuration for kafka topic is not applied in this repo at all.

style95: The `scheduling` configurations are required.

```
      "CONFIG_whisk_scheduler_scheduling_staleThreshold": "{{ scheduler.scheduling.staleThreshold }}"
      "CONFIG_whisk_scheduler_scheduling_checkInterval": "{{ scheduler.scheduling.checkInterval }}"
      "CONFIG_whisk_scheduler_scheduling_dropInterval": "{{ scheduler.scheduling.dropInterval }}"
```
https://github.com/apache/openwhisk/blob/master/ansible/roles/schedulers/tasks/deploy.yml#L117

style95: `CONFIG_whisk_scheduler_queue_maxBlackboxRetentionMs` is required.
https://github.com/apache/openwhisk/blob/master/ansible/roles/schedulers/tasks/deploy.yml#L128

helm/openwhisk/templates/_helpers.tpl
```diff
@@ -224,12 +238,20 @@ app: {{ template "openwhisk.fullname" . }}
...
   value: {{ .Values.whisk.kafka.topics.invoker.retentionMs | quote }}
 - name: "CONFIG_whisk_kafka_topics_invoker_segmentBytes"
   value: {{ .Values.whisk.kafka.topics.invoker.segmentBytes | quote }}
+
+- name: "CONFIG_whisk_kafka_topics_scheduler_retentionBytes"
+  value: {{ .Values.whisk.kafka.topics.scheduler.retentionBytes | quote }}
+- name: "CONFIG_whisk_kafka_topics_scheduler_retentionMs"
+  value: {{ .Values.whisk.kafka.topics.scheduler.retentionMs | quote }}
+- name: "CONFIG_whisk_kafka_topics_scheduler_segmentBytes"
+  value: {{ .Values.whisk.kafka.topics.scheduler.segmentBytes | quote }}
```
style95, line 254: Not only topic configurations for the `scheduler` topic but also for the `creationAck` topic are required.
https://github.com/apache/openwhisk/blob/master/ansible/roles/schedulers/tasks/deploy.yml#L144

style95 (commented):
I am a bit outdated regarding this repo as I am using a bit different chart and scripts to deploy ow on k8s in my downstream.
But anyway I think at least the followings are required.

style95:
@hunhoffe 
This is great! 
Thank you for your contribution.

helm/openwhisk/templates/scheduler-pod.yaml
hunhoffe: Regarding ```whisk_cluster_name```: I believe you name a deployment when you create it with helm (which becomes ```Release.name``` in the helm template files), which would allow you to deploy multiple clusters without conflict (I am speaking hypothetically, as I have not tried this myself). If I am understanding correctly, perhaps ```whisk_cluster_name``` is not needed here. @dgrove-oss does this seem correct?

style95: The cluster name is used to differentiate multiple versions when two different releases share the same ETCD.
https://github.com/apache/openwhisk/blob/master/common/scala/src/main/scala/org/apache/openwhisk/core/etcd/EtcdUtils.scala#L72

In this change, it seems each release uses its own ETCD but when an external ETCD is shared among multiple clusters for some reason(some may want to deploy clusters in a blue-green way), the config would be required.

helm/openwhisk/values.yaml
hunhoffe: I've added a scheduler check to the travis test matrix

helm/openwhisk/templates/scheduler-pod.yaml
hunhoffe: I changed to ```waitForKafka```, ```waitForCouchDB```, and ```waitForEtcd```, and also added ```waitForEtcd``` to the controller.

helm/openwhisk/templates/scheduler-pod.yaml
hunhoffe: I have some confusion with how ```whisk cluster name``` is implemented in the main OpenWhisk repo, and I'd appreciate your help in understanding why it is the way it is.

In the link you pasted it is ```CONFIG_whisk_cluster_name```, but in config hierarchy it is ```whisk.cluster_name```. My understanding is that it should be ```CONFIG_whisk_clusterName``` to follow convention.

To my understanding it is etcd specific, so shouldn't it be under ```etcd.clusterName``` and not in ```whisk.clusterName```? I believe elasticsearch config takes this approach here: https://github.com/apache/openwhisk/blob/a1639f0e4d7270c9a230190ac26acb61413b6bbb/ansible/roles/elasticsearch/templates/elasticsearch.yml.j2#L1

I am asking these questions because I don't want to modify the main whisk config for scheduler-specific things if I don't have to. If there is a reason for it to be the way it is currently in OpenWhisk, I can still add it to this PR. But I wanted some clarification first.

style95: It is not only being used by schedulers but also by controllers too.
https://github.com/apache/openwhisk/blob/1fbfd6bcb3d65845a328382641c1db7182763ced/core/controller/src/main/scala/org/apache/openwhisk/core/loadBalancer/FPCPoolBalancer.scala#L72

hunhoffe:
I believe I've addressed all comments except the external etcd feature. I wrote a comment that it is not yet supported in the values.yaml file, and I'm hoping that it is a feature that could be added later; I could create an issue for it so it is not forgotten.

helm/openwhisk/values.yaml
```diff
@@ -139,9 +139,9 @@ whisk:
     includeSystemTests: false
   versions:
     openwhisk:
-      buildDate: "2022-06-23-06:18:00Z"
```
hunhoffe, line 131: I made the details of this timestamp up. Am I supposed to have copied it from somewhere?

helm/openwhisk/templates/_helpers.tpl
style95: Can we add the same configurations for the `creationAck` topic too?

helm/openwhisk/templates/scheduler-pod.yaml
```diff
@@ -0,0 +1,273 @@
...
+        - name: "POD_IP"
+          valueFrom:
+            fieldRef:
+              fieldPath: status.podIP
+
+        - name: "WHISK_SCHEDULER_ENDPOINTS_RPCPORT"
+          value: {{ .Values.scheduler.rpcPort | quote }}
+
+        - name: "WHISK_SCHEDULER_ENDPOINTS_AKKAPORT"
+          value: "25520"
```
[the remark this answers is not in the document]
style95: Seems this is still not configurable.

style95:
> I could create an issue for it so it is not forgotten if that's acceptable.
I am ok with this.

helm/openwhisk/values.yaml
style95: I believe @dgrove-oss has an idea.

dgrove-oss: My habit has been to use the UTC time of the commit we are using in `gitTag` as the `buildDate`.  I don't think this actually matters, but it's what I'd been doing ;)

style95 (approved):
LGTM

PR #733, opened 2022-07-11
Add permission to the events

style95:
This is the counterpart of this PR https://github.com/apache/openwhisk/pull/5270

### DIFF
--- a/helm/openwhisk/templates/invoker-role.yaml
+++ b/helm/openwhisk/templates/invoker-role.yaml
@@ -35,6 +35,9 @@ rules:
   resources: ["pods"]
   verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
 - apiGroups: [""]
+  resources: ["events"]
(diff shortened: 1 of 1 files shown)


helm/openwhisk/templates/invoker-role.yaml
```diff
@@ -34,6 +34,9 @@ rules:
 - apiGroups: [""]
   resources: ["pods"]
   verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+- apiGroups: [""]
+  resources: ["events"]
+  verbs: ["get", "list"]
```
ningyougang, line 39: Seems our downstream doesn't have this changes,
So we changed the events k8s resource's permission manually?

ningyougang: Our downstream use another invoker-role.yaml which already applied above configuration.

ningyougang (approved):

PR #741, opened 2022-08-11
Add description for PVC with static volumes

### LINKED ISSUE #740: PVC does not get fulfilled
For OpenWhisk to run, I have created a StorageClass with the provisioner `kubernetes.io/no-provisioner`.  
Because I don't have dynamic provisioning available, I created six PeristantVolumes of size 2Gi.  

When I switched my setup and decreased the PeristantVolume size from 2Gi to 1Gi, my setup would fail, because the PVC of CouchDB would not be fulfilled.  
As it turns out, the PVC of CouchDB requires [by default 2Gi](https://github.com/apache/openwhisk-deploy-kube/blob/4ec933c557dde851d92eadd0dcf5480af64f3cce/helm/openwhisk/values.yaml#L227), so increasing the size again solved the problem.  

Should we add a small note for the volume size [here](https://github.com/apache/openwhisk-deploy-kube/tree/master/helm/openwhisk) or are people expected to use dynamic provisioning where this is no issue?

linuswagner:
Closes #740


dgrove-oss:
Thank you @linuswagner

PR #754, opened 2022-10-20
Prevent etcd from running when scheduler disabled

hunhoffe:
I recently noticed that etcd was running even when the scheduler is disabled. This PR exists primarily to fix this bug.

However, since I was already making a PR, I also briefly tested deployment with the most recent OpenWhisk images, and updated them here (there have been some updates to the scheduler since the last image was accepted here).


dgrove-oss:
Thanks @hunhoffe!

dgrove-oss (approved):

Issue #735, opened 2022-07-15
Install-package failed when deploying from helm repo


lyudmilalala:
I am trying to deploy Openwhisk on Windows Docker Desktop k8s

If I following the [Deploying Released Charts from Helm Repository Section in README](https://github.com/apache/openwhisk-deploy-kube/blob/master/README.md#deploy-with-helm), the install-package pod will finally get status `Error`. 

Final status of pods looks like this
```
NAMESPACE     NAME                                     READY   STATUS      RESTARTS          AGE
openwhisk     owdev-alarmprovider-687f79859b-rnhxk     1/1     Running     0                 5h46m
openwhisk     owdev-apigateway-56b8cc9dbf-kxbdz        1/1     Running     0                 5h46m
openwhisk     owdev-controller-0                       1/1     Running     0                 5h46m
openwhisk     owdev-couchdb-59f57cbff7-rxhkt           1/1     Running     0                 5h46m
openwhisk     owdev-gen-certs--1-6gtl7                 0/1     Completed   0                 5h46m
openwhisk     owdev-init-couchdb--1-gh9jl              0/1     Completed   0                 5h46m
openwhisk     owdev-install-packages--1-4wjvq          0/1     Error       0                 5h46m
openwhisk     owdev-install-packages--1-nh2hw          0/1     Error       0                 5h12m
openwhisk     owdev-install-packages--1-npqgd          0/1     Error       0                 5h16m
openwhisk     owdev-install-packages--1-pxn47          0/1     Error       0                 5h23m
openwhisk     owdev-invoker-0                          1/1     Running     0                 5h46m
openwhisk     owdev-kafka-0                            1/1     Running     0                 5h46m
openwhisk     owdev-kafkaprovider-5574d4bf5f-spfkd     1/1     Running     0                 5h46m
openwhisk     owdev-nginx-74fdfbf868-czfhk             1/1     Running     0                 5h46m
openwhisk     owdev-redis-68bfb4ccb8-n5ppj             1/1     Running     0                 5h46m
openwhisk     owdev-wskadmin                           1/1     Running     0                 5h46m
openwhisk     owdev-zookeeper-0                        1/1     Running     0                 5h46m
openwhisk     wskowdev-invoker-00-1-prewarm-nodejs10   1/1     Running     0                 5h44m
openwhisk     wskowdev-invoker-00-2-prewarm-nodejs10   1/1     Running     0                 5h44m
```

[...]

PR #772, opened 2023-12-11
Add warning about the old Helm charts

agcom:
Related to #735; let's save people (like me) from wasting time debugging their deployment.


agcom:
Are there any plans to update the Helm charts at the Helm repository?

PR #783, opened 2024-07-31
fix: etcd-pvc being deployed even etcd-pod not being deployed

### LINKED ISSUE #782: Orphan PVC created
We have noticed an instance of an orphan resource in one of your Kubernetes configuration files. Deploying Kubernetes resources that are not being actively used or referenced by any other resources in the cluster can lead to orphan resources. These orphan resources consume unnecessary resources and can potentially cause confusion or clutter within the cluster.  Additionally, we provide anecdotal evidence from [topolvm-topolvm#484](https://github.com/topolvm/topolvm/issues/484) regarding the orphan resource defect.

In etcd-pod.yaml, we can see only when scheduler.enabled = true and etcd.external = false the etcd pod can be deployed.

https://github.com/apache/openwhisk-deploy-kube/blob/5a5a2168c9c0342c026ca137f96fa9d2c8ae2748/helm/openwhisk/templates/etcd-pod.yaml#L18

But in etcd-pvc.yaml, When etcd.external = false and k8s.persistence.enabled = true, the etcd pvc will be deployed.

https://github.com/apache/openwhisk-deploy-kube/blob/5a5a2168c9c0342c026ca137f96fa9d2c8ae2748/helm/openwhisk/templates/etcd-pvc.yaml#L18

However, in the values.yaml file we can see that the default value for scheduler.enabled is false, for etcd.external is false, and for k8s.persistence.enabled is true. That means etcd pod will not be deployed but the relevant PVC will be deployed. And this PVC not being actively used or referenced by any other resources, it is a orphan resource.

https://github.com/apache/openwhisk-deploy-kube/blob/5a5a2168c9c0342c026ca137f96fa9d2c8ae2748/helm/openwhisk/values.yaml#L274-L275

https://github.com/apache/openwhisk-deploy-kube/blob/5a5a2168c9c0342c026ca137f96fa9d2c8ae2748/helm/openwhisk/values.yaml#L315-L317

https://github.com/apache/openwhisk-deploy-kube/blob/5a5a2168c9c0342c026ca137f96fa9d2c8ae2748/helm/openwhisk/values.yaml#L157-L158

## Expected Behavior

etcd-pvc not being deployed if etcd pod not being deployed.

zyue110026:
Fixes #782

### DIFF
--- a/helm/openwhisk/templates/etcd-pvc.yaml
+++ b/helm/openwhisk/templates/etcd-pvc.yaml
@@ -15,7 +15,7 @@
 # limitations under the License.
 #
 
-{{- if and (not .Values.etcd.external) .Values.k8s.persistence.enabled }}
+{{- if and (not .Values.etcd.external) .Values.k8s.persistence.enabled (.Values.scheduler.enabled) }}
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:


dgrove-oss (approved):
Thanks for the contribution.

